PR #9, opened 2019-08-23
Provide human-readable timestamps in changes_timestamps check

leplatrem:
It's very often that I need to check the datetime when a collection was changed. Sometimes I use the notification emails because the timestamps in the `/buckets/monitor/collections/changes` are not human readable.

It would be useful to have human-readable at hand always (from CLI or UI). Hence this changes that returns them always as data, even if the check returns true.

### DIFF
--- a/checks/remotesettings/changes_timestamps.py
+++ b/checks/remotesettings/changes_timestamps.py
@@ -31,21 +31,20 @@ async def run(request, server):
         for entry in entries
     ]
     results = await asyncio.gather(*futures)
-    all_good = True
+
     datetimes = []
-    for (entry, timestamp) in zip(entries, results):
-        collection_timestamp = timestamp
-        entry_timestamp = str(entry["last_modified"])
-        all_good = all_good and collection_timestamp == entry_timestamp
-        dt = datetime.utcfromtimestamp(int(timestamp) / 1000).isoformat()
-        cid = "{bucket}/{collection}".format(**entry)
+    for (entry, collection_timestamp) in zip(entries, results):
+        entry_timestamp = entry["last_modified"]
+        collection_timestamp = int(collection_timestamp)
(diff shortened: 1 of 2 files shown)


leplatrem:
@glasserc r?

checks/remotesettings/changes_timestamps.py
```diff
@@ -30,10 +31,21 @@ def get_timestamp(client, bucket, collection, timestamp):
         for entry in entries
     ]
     results = await asyncio.gather(*futures)
-    failing = []
+    all_good = True
+    datetimes = []
     for (entry, timestamp) in zip(entries, results):
+        collection_timestamp = timestamp
```
glasserc: Why not just assign `collection_timestamp` in the `for` statement?

checks/remotesettings/changes_timestamps.py
```diff
@@ -30,10 +31,21 @@ def get_timestamp(client, bucket, collection, timestamp):
...
+        entry_timestamp = str(entry["last_modified"])
```
glasserc: Why should we convert these to strings? The fact that we're parsing them as millis-since-the-epoch means that they essentially "have" to be integers..

checks/remotesettings/changes_timestamps.py
```diff
@@ -30,10 +31,21 @@ def get_timestamp(client, bucket, collection, timestamp):
...
+        all_good = all_good and collection_timestamp == entry_timestamp
```
glasserc: This is a little fancy. I'd rather see `all_good` be calculated declaratively outside the for loop.

checks/remotesettings/changes_timestamps.py
leplatrem: Good point, thanks!

PR #15, opened 2019-08-28
Make CI pass

mythmon:
Before this change, CI doesn't work and I couldn't run the project. I think this change will fix both problems.


.flake8
```diff
@@ -0,0 +1,2 @@
+[flake8]
+ignore=W503,E501
```
leplatrem: no max line length ?

.circleci/config.yml
```diff
@@ -10,17 +10,12 @@ jobs:
...
       - run:
           name: Create version.json
-          command: |
-            printf '{"commit":"%s","version":"%s","source":"https://github.com/%s/%s","build":"%s"}\n' \
-            "$CIRCLE_SHA1" \
-            "$CIRCLE_TAG" \
-            "$CIRCLE_PROJECT_USERNAME" \
-            "$CIRCLE_PROJECT_REPONAME" \
-            "$CIRCLE_BUILD_URL" | tee version.json
+          command: make version.json
```
leplatrem: Maybe we should use those `$CIRCLE_*` variables in the Makefile if we do that

Makefile
```diff
@@ -13,10 +13,10 @@ INSTALL_STAMP := $(VENV)/.install.stamp
 .PHONY: clean check lint format tests
 
 install: $(INSTALL_STAMP) $(COMMIT_HOOK)
-$(INSTALL_STAMP): $(PYTHON) dev-requirements.txt requirements.txt checks/remotesettings/requirements.txt
-	$(VENV)/bin/pip install -Ur requirements.txt
-	$(VENV)/bin/pip install --no-deps -Ur checks/remotesettings/requirements.txt
-	$(VENV)/bin/pip install -Ur dev-requirements.txt
+$(INSTALL_STAMP): $(PYTHON) requirements/dev.txt requirements/default.txt checks/remotesettings/requirements.txt
+	$(VENV)/bin/pip install -Ur requirements/default.txt
+	$(VENV)/bin/pip install -Ur checks/remotesettings/requirements.txt
```
leplatrem: We want the `--no-deps` here because of kinto-signer. Otherwise it will pull the whole pyramid ecosystem :) I should have added a comment.

.flake8
mythmon: Black formats lines with a soft limit of 88 characters, but sometimes will produce lines slightly longer than that. I've found it best to let Black handle all formatting questions, including line length, and let flake8 focus on structural concerns.

Makefile
mythmon: Ah, I wondered why Pyramid was being pulled in. I'll put `--no-deps` in and re-do the constraints file.

I wonder if it would be possible to add some variant of `kinto-signer` like `kinto-signer[no-web]` that doesn't rely on Pyramid?

.circleci/config.yml
mythmon: I couldn't figure out how to sanely use those `$CIRCLE_` variables in the makefile, so I put this back.

Makefile
leplatrem: > I wonder if it would be possible to add some variant of kinto-signer like kinto-signer[no-web] that doesn't rely on Pyramid?

I know that the opposite is possible (extra deps) but not sure if `[no-deps]` is feasible. The best would be to extract the utils of kinto-signer into an external package...

PR #43, opened 2019-09-13
Fix #42: Refactor conf and query params

### LINKED ISSUE #42: Use query params in cache key
And probably whitelist them

leplatrem:
Fix #42


poucave/main.py
```diff
@@ -47,41 +48,74 @@ def __init__(self):
...
+                raise web.HTTPBadRequest()
+
+            # Some parameters are exposed in JSON response.
+            final_params = {k: v for k, v in params.items() if k in exposed_params}
+
             # Each check has its own TTL.
-            cache_key = f"{project}/{name}"
+            cache_key = f"{project}/{name}-" + ",".join(
+                f"{k}:{v}" for k, v in params.items()
+            )
```
glasserc: This uses the parameters from the conf as well. I think that's probably correct -- it means we ignore the cache if we change parameters -- but it also could mean that we have cached results from a previous version of the application if we restart with the same configuration.

leplatrem: Yes, the thing is that the cache is in memory, so not persistent between restarts

PR #45, opened 2019-09-13
Fix #34: Check for uptake error rate

### LINKED ISSUE #34: Add a check for error rate in clients uptake Telemetry
Similar to what Peter did in https://github.com/mozilla-services/remote-settings-lambdas/blob/master/commands/uptake_health.py

leplatrem:
Fix #34

* [ ] ~~Parametrize Redash query~~ ...troublesome with result caching/scheduling. Hard-coding `6 hours` and `component = 'remotesettings'`  for now 
* [x] Add min total entries param
* [x] Document everything in query https://sql.telemetry.mozilla.org/queries/64808/source
* [ ] ~~Retry failed HTTP request~~ --> Will be done in different request
* [ ] ~~Fetch last X hours~~ (cf. parametrize)
* [x] Show min/max date
* [x] Expose params
* [x] Tests

Screenshot from 2019-09-13 16-54-24


checks/remotesettings/clients_error_rate.py
```diff
@@ -0,0 +1,125 @@
...
+EXPOSED_PARAMETERS = ["max_percentage", "min_total_events"]
+
+REDASH_URI = (
+    f"https://sql.telemetry.mozilla.org/api/queries/64808/results.json?api_key="
+)
+REQUESTS_TIMEOUT_SECONDS = int(os.getenv("REQUESTS_TIMEOUT_SECONDS", 5))
+
+
+def sort_dict_desc(d, key):
+    return dict(sorted(d.items(), key=key, reverse=True))
```
glasserc: Can we use `OrderedDict` here?

tests/checks/remotesettings/test_clients_error_rate.py
```diff
@@ -0,0 +1,127 @@
...
+        "min_timestamp": "2019-09-16T01:36:12.348",
+        "max_timestamp": "2019-09-16T07:24:58.741",
+    }
+
+
+async def test_min_total_events():
+    with mock.patch(f"{MODULE}.fetch_redash") as mocked:
+        f = asyncio.Future()
+        f.set_result(FAKE_ROWS)
+        mocked.return_value = f
```
glasserc: It kinda feels like this mock should be refactored to be more DRY.

checks/remotesettings/clients_error_rate.py
```diff
@@ -0,0 +1,125 @@
...
+
+    query_result = body["query_result"]
+    data = query_result["data"]
+    rows = data["rows"]
+    return rows
+
+
+async def run(
+    api_key: str,
+    max_percentage: float,
```
glasserc: Instead of `max_percentage`, how about `max_error_percentage`?

checks/remotesettings/clients_error_rate.py
```diff
@@ -0,0 +1,125 @@
...
+
+import aiohttp
+
+from poucave.typings import CheckResult
+
+
+EXPOSED_PARAMETERS = ["max_percentage", "min_total_events"]
+
+REDASH_URI = (
+    f"https://sql.telemetry.mozilla.org/api/queries/64808/results.json?api_key="
```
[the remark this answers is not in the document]
leplatrem: I thought Redash was the way to go (JSON, API Key, periodic schedule...).

You can review the query, you're allowed (and most of our team mates) to view/edit it.

I can investigate about BigQuery, but for what I can see it will be more painful in terms of authentication, etc.

checks/remotesettings/clients_error_rate.py
leplatrem: dict are sorted in 3.6+, so I don't think using OrderedDict it's necessary

checks/remotesettings/clients_error_rate.py
```diff
@@ -0,0 +1,125 @@
...
+    ignore_status: List[str] = [],
+) -> CheckResult:
+    rows = await fetch_redash(api_key)
+
+    min_timestamp = "9999"
+    max_timestamp = "0000"
+    by_collection: Dict[str, Dict[str, int]] = defaultdict(dict)
+    for row in rows:
+        if row["min_timestamp"] < min_timestamp:
+            min_timestamp = row["min_timestamp"]
```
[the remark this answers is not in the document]
leplatrem: Thanks!

checks/remotesettings/clients_error_rate.py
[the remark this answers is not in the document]
glasserc: We spoke online about this. The query takes minutes to run, which makes running it directly from a request problematic. Using Redash allows us to schedule the query so that when we make requests, we get results immediately. On the other hand, it fragments the code for this service, and it introduces a weird kind of bus factor since the query is under Mat's account. We agreed that using Redash was fine for now but we would introduce an issue to follow up and see if there's a better way to address this (#60).

PR #47, opened 2019-09-16
Fix #11: Add check for certificates expiration

### LINKED ISSUE #11: Add a check about certificates expiring soon

leplatrem:
Fix #11


checks/remotesettings/certificates_expiration.py
```diff
@@ -0,0 +1,86 @@
+"""
+Signature certificates should expire in a minimum number of days.
+
+The list of collections whose certificate expires too soon is returned, along
+with the expiration date and x5u URL.
```
glasserc: I think this wording is a bit awkward. How about something like:

```suggestion
Signature certificates should not expire for at least some minimum number of days.

Returns a list of collections whose certificate expires too soon, along with their expiration date and x5u URL.
```

checks/remotesettings/certificates_expiration.py
```diff
@@ -0,0 +1,86 @@
...
+        {
+            metadata["signature"]["x5u"]
+            for metadata in results
+            if "signature" in metadata
+        }
+    )
+    futures = [
+        loop.run_in_executor(None, fetch_certificate_expiration, x5u) for x5u in x5us
+    ]
+    results = await asyncio.gather(*futures)
```
glasserc: The fact that we have more than one `asyncio.gather` seems a little bit like a code smell to me. At first I thought this was an optimization because some certificates are repeated, but it seems like we don't do that either. (If we do, then we should put a comment, something like `Gather x5us first to deduplicate`.)

It seems to me that we don't need the intermediate result. Could this maybe be cleaner if we chain the futures in some other way, like a function that returns the URL and its expiration as a pair?

checks/remotesettings/certificates_expiration.py
```diff
@@ -0,0 +1,86 @@
...
+    expirations = {x5u: expiration for x5u, expiration in zip(x5us, results)}
+
+    errors = {}
+    for entry, metadata in entries_metadata:
+        cid = "{bucket}/{collection}".format(**entry)
+        try:
+            x5u = metadata["signature"]["x5u"]
+        except KeyError:
+            errors[cid] = {"x5u": None, "expires": None}
+            continue
```
glasserc: This `try` seems a bit excessive. We don't use `try` around `fetch_certificate_expirations`, for instance (which I think is good and correct, because we don't expect that fetch to fail or produce bad data). Besides, it kind of conflates "doesn't have a signature" and "signature doesn't have a certificate". Maybe we should distinguish these?

leplatrem: I agree!

We have other checks that will fail nicely about this missing signature issue. I removed the safety check here.

checks/remotesettings/certificates_expiration.py
leplatrem: Yes we do deduplicate (the set syntax is not super visible), I added a comment.

glasserc: OK, thanks, I don't know why I didn't see it the first time through.

PR #49, opened 2019-09-16
Fix #27: check deployed version

### LINKED ISSUE #27: Add check for deployed version
A check that would take a Github repo URL, a service URL, and a «*max lag*» parameters, in order to compare the latest Github tag with the deployed version (`/__version__`).

leplatrem:
Fix #27

### DIFF
--- a/checks/core/deployed_version.py
+++ b/checks/core/deployed_version.py
@@ -1,0 +1,37 @@
+"""
+The deployed `version` should be the latest tag of the specified `repo`.
+
+The deployed version and the latest tag are returned.
+"""
+import os
+
+import aiohttp
+
+from poucave.typings import CheckResult
+
+
+EXPOSED_PARAMETERS = ["server", "repo"]
+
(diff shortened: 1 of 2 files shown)


checks/core/deployed_version.py
```diff
@@ -0,0 +1,37 @@
...
+        async with session.get(
+            f"https://api.github.com/repos/{repo}/releases"
+        ) as response:
+            releases = await response.json()
+
+    deployed_version = version_info["version"]
+    latest_tag = releases[0]["tag_name"]
+
+    return (
+        deployed_version == latest_tag,
```
glasserc: This might be a nit, but can we also allow `v{deployed_version}`? That's how we tag e.g. kinto-http.js. https://github.com/Kinto/kinto-http.js/tree/v4.7.3

leplatrem: Yeah, but I think we use the Github tag to produce the `version.json` so we should be safe

See https://github.com/mozilla/normandy/blob/master/.circleci/config.yml#L69

PR #52, opened 2019-09-17
Unify responses and aioresponses

leplatrem:
Previously you had to `import responses` to register mock responses because you had to have access to `responses.HEAD` and `responses.GET`. Additionally, it was hard to remember if it was `mock_` or `mocked_` and `response` or `responses`. Harmonize this for `responses` and `aioresponses`.

### DIFF
--- a/tests/checks/core/test_deployed_version.py
+++ b/tests/checks/core/test_deployed_version.py
@@ -1,9 +1,9 @@
 from checks.core.deployed_version import run
 
 
-async def test_positive(mock_aioresponse):
+async def test_positive(mock_aioresponses):
     url = "http://server.local/v1"
-    mock_aioresponse.get(
+    mock_aioresponses.get(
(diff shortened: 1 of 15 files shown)


leplatrem:
@glasserc do you think we should have a fixture that mocks both requests/aiohttp ?

leplatrem:
As discussed on Zoom, we wanted to have a single `mock_responses` fixture, that would mock both `aiohttp` and `requests`.

Unfortunately this does not seem to be possible, since `aioresponses` will raise an `AssertionError` if not all mocked responses were consumed...

```
E           AssertionError: Not all requests have been executed [('GET', 'http://n/api/v1/recipe/signed/'), ('GET', 'http://rs/v1/buckets/main/collections/normandy-recipes/records')]

.venv/lib/python3.7/site-packages/responses.py:643: AssertionError

```

PR #71, opened 2019-09-20
Limit parallel requests everywhere

leplatrem:


leplatrem:
I realize that we could have a `run_parallel()` util here. Especially because we have a lot of other places where we do parallel requests (everywhere `asyncio.gather(*futures)` is used almost)

leplatrem:
> It would also give us a mechanism to hide the zip everywhere (which I think is a little ugly).

I don't really know how to do it better. Do you have ideas?

leplatrem:
> I guess you didn't like the previous approach you took, run_parallel(f, [args1, args2])?

Yeah, I think I prefer to see explicitly the actual call. Plus it does not really solve our `zip()` situation :)

PR #76, opened 2019-09-20
Fix Normandy uptake ignore backoff

leplatrem:
This is kinda crazy. The tests were passing because the wrong line was mutating the function parameters :/

### DIFF
--- a/checks/normandy/uptake_error_rate.py
+++ b/checks/normandy/uptake_error_rate.py
@@ -41,9 +41,8 @@ async def run(
     ignore_status: List[str] = [],
 ) -> CheckResult:
     # Ignored statuses are specified using the Normandy ones.
-    ignored_status = [UPTAKE_STATUSES.get(s, s) for s in ignore_status]
-    # A client reporting that recipe didn't match filter is not an error.
-    ignore_status += ["recipe_didnt_match_filter"]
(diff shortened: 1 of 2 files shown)


checks/normandy/uptake_error_rate.py
```diff
@@ -41,9 +41,8 @@ def sort_dict_desc(d, key):
     ignore_status: List[str] = [],
 ) -> CheckResult:
     # Ignored statuses are specified using the Normandy ones.
-    ignored_status = [UPTAKE_STATUSES.get(s, s) for s in ignore_status]
-    # A client reporting that recipe didn't match filter is not an error.
-    ignore_status += ["recipe_didnt_match_filter"]
+    # A client reporting that recipe didn't match filter (backoff) is not an error.
+    ignored_status = [UPTAKE_STATUSES.get(s, s) for s in ignore_status + ["backoff"]]
```
glasserc: I still don't see what the significance of this addition is. Without this addition, `total_statuses` equals the sum of all statuses, `total_errors` does not include `backoff`, and `error_rate` is `total_errors / total_statuses`. With this addition, `total_statuses` is ... the same? `total_errors` is ... the same?

leplatrem: Hmm. You are absolutely right. `backoff` does not end with `_error` :medal_sports:

PR #77, opened 2019-09-20
Reach 100% of coverage for checks

leplatrem:


checks/remotesettings/utils.py
```diff
@@ -17,7 +17,7 @@
 )
 
 
-class CustomTimeout(TimeoutSauce):
+class CustomTimeout(TimeoutSauce):  # pragma: no cover
```
leplatrem: I realized this is not used at all. The new way to go will be https://github.com/Kinto/kinto-http.py/pull/216

PR #92, opened 2019-09-25
Fix #86: Use x5u to validate signature 

### LINKED ISSUE #86: Use x5u instead of public_key to validate signatures

leplatrem:
Fix #86


leplatrem:
> LGTM. Is there anything I should be doing to try to examine the certificates?

Examining the certificate... hmm, thanks! You helped me to think about datetime comparisons... if we freeze the certificate in the tests, we should also freeze the datetime comparisons.

PR #93, opened 2019-09-26
Fix #82: option to ignore client versions

### LINKED ISSUE #82: Improve uptake error rate to discard unsupported versions of clients
Some old Firefox versions report errors that have been fixed in our latest releases.

We should investigate if there is a way to:
* obtain the list of supported versions (bedrock?)
* filter out versions

The client version can be obtained with this:
```
SELECT application.display_version, application.version
FROM `moz-fx-data-shared-prod.telemetry_live.event_v4`
WHERE date(submission_timestamp)='2019-09-19'
LIMIT 10
```

leplatrem:
Fix #82

### DIFF
--- a/checks/remotesettings/uptake_error_rate.py
+++ b/checks/remotesettings/uptake_error_rate.py
@@ -13,9 +13,14 @@ from poucave.typings import CheckResult
 from poucave.utils import fetch_redash
 
 
-EXPOSED_PARAMETERS = ["max_error_percentage", "min_total_events"]
+EXPOSED_PARAMETERS = [
+    "max_error_percentage",
+    "min_total_events",
+    "ignore_status",
+    "ignore_versions",
+]
 
-REDASH_QUERY_ID = 64808
+REDASH_QUERY_ID = 65039
 
 
 def sort_dict_desc(d, key):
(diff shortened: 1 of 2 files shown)


checks/remotesettings/uptake_error_rate.py
```diff
@@ -27,36 +32,39 @@ def sort_dict_desc(d, key):
     max_error_percentage: float,
     min_total_events: int = 10000,
     ignore_status: List[str] = [],
+    ignore_versions: List[int] = [],
```
glasserc: Should versions be an `int`? What about 70.1 or 68.0.1?

leplatrem: Ok, you're right, that is worth discussing.

I do the `SPLIT(version, ".")[0]` in BigQuery. We could do it here instead. But I supposed that we would ignore major ones only, and this would reduce the amount of data to fetch from the Redash JSON API.

I'm happy to change if you have a strong opinion towards strings, or even regex or whatever

glasserc: No, that's fine, thanks.

PR #99, opened 2019-09-27
Fix #90: Uptake duration percentiles

### LINKED ISSUE #90: Check .95 percentile duration of success syncs is less than X seconds

leplatrem:
Fix #90

* [x] Basic concept
* [x] Refactor/document Redash query
* [x] Docstrings
* [x] Add tests

### DIFF
--- a/checks/remotesettings/uptake_max_duration.py
+++ b/checks/remotesettings/uptake_max_duration.py
@@ -25,6 +25,8 @@ async def run(
 
     duration_percentiles = rows[0]["duration_percentiles"]
 
+    # Percentiles have `str` type because config keys are strings in TOML.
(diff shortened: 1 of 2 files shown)


checks/remotesettings/uptake_max_duration.py
```diff
@@ -0,0 +1,58 @@
...
+from typing import Dict
+
+from poucave.typings import CheckResult
+from poucave.utils import fetch_redash
+
+REDASH_QUERY_ID = 65069
+
+
+async def run(
+    api_key: str, max_percentiles: Dict[str, int], source: str = "settings-sync"
```
glasserc: Why encode percentiles as `str`?

PR #100, opened 2019-09-27
Fix #91: Uptake age percentiles

### LINKED ISSUE #91: Check .95 percentile age of data of success sync is less than X seconds

leplatrem:
Fix #91

* [x] Basic concept
* [x] Refactor/document query
* [x] Add docstrings
* [x] Add tests

### DIFF
--- a/checks/remotesettings/uptake_max_age.py
+++ b/checks/remotesettings/uptake_max_age.py
@@ -1,0 +1,54 @@
+"""
+"""
+from typing import Dict
+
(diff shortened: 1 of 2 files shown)


leplatrem:
First I was using the same query and the same code, but the little differences (ability to pick the source in one, the filter on `source=trigger` in the other, etc.) made me change my mind. They are similar but also very short and simple, so reusing was not especially making things look simpler...

PR #120, opened 2019-10-11
Add more details about collection inconsistencies

leplatrem:
Do you think we should also return the records IDs?


checks/remotesettings/collections_consistency.py
```diff
@@ -24,15 +24,25 @@ def records_equal(a, b):
...
+    extras = len(b_by_id)
+
+    msg = ""
+    if missing > 0:
+        msg += f"{missing} records missing. "
+    if differ > 0:
+        msg += f"{differ} records differ. "
+    if extras > 0:
+        msg += f"{extras} records extra. "
+    return msg
```
glasserc: Maybe use `" ".join()` instead?

Issue #123, opened 2019-10-15
Container won't start


bqbn:
Using mozilla/poucave:1.3.1 image, the container won't start with the following error:

```
$ kubectl logs stage-poucave-app-1-deploy-6789b75db6-kzl5v poucave
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "poucave/__main__.py", line 5, in <module>
    main(sys.argv[1:])
  File "/app/poucave/main.py", line 208, in main
    app = init_app(conf)
  File "/app/poucave/main.py", line 149, in init_app
    handler = handlers.checkpoint(project, check, **params)
  File "/app/poucave/main.py", line 60, in checkpoint
    mod = importlib.import_module(module)
  File "/usr/local/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/app/checks/remotesettings/validate_signatures.py", line 14, in <module>
    import cryptography.x509
  File "/usr/local/lib/python3.7/site-packages/cryptography/x509/__init__.py", line 7, in <module>
    from cryptography.x509 import certificate_transparency
  File "/usr/local/lib/python3.7/site-packages/cryptography/x509/certificate_transparency.py", line 10, in <module>
    import six
ModuleNotFoundError: No module named 'six'
```

It looks like the `six` module is missing inside the image.

leplatrem:
Wao, that's weird, since we run the tests from the container on each commit.
And I can see on CircleCI that `six` gets installed. https://circleci.com/gh/mozilla-services/poucave/313#tests/containers/0

I'll investigate...

leplatrem:
The situation is the following:
[...]

PR #125, opened 2019-10-15
Test that Docker runs successfully

leplatrem:
In order to avoid facing #123 again. I'd like to run the container with some remotesettings checks and get its heartbeat. 

Mounting the config won't work :( @glasserc do you have an idea?


leplatrem:
I do exactly this! https://circleci.com/docs/2.0/building-docker-images/#accessing-services

Dockerfile
```diff
@@ -22,6 +22,9 @@ RUN pip install --progress-bar=off -r requirements/default.txt && \
 
 COPY . /app
 
+# File must exist otherwise volume is mounted as directory.
+RUN touch /app/config.toml
```
glasserc: I don't really understand this. Isn't `/app` being mounted as directory what we want? Were you able to get it to work mounting just a file?

leplatrem: You're right, I got confused. It's when the source does not exist that a directory gets created.

PR #136, opened 2019-10-21
Fix #105: Include checks about normandy-recipes-capabilities

### LINKED ISSUE #105: Improve Normandy check to include `normandy-recipes-capabilities`
See https://github.com/mozilla/normandy/pull/1960

leplatrem:
Fix #105

### DIFF
--- a/checks/normandy/remotesettings_recipes.py
+++ b/checks/normandy/remotesettings_recipes.py
@@ -8,45 +8,68 @@ inconsistencies between the baseline and capabilities collections.
 from poucave.typings import CheckResult
 from poucave.utils import fetch_json
 
-NORMANDY_URL = "{server}/api/v1/recipe/signed/?enabled=1"
+NORMANDY_URL = "{server}/api/v1/recipe/signed/?enabled=1&only_baseline_capabilities={baseline_only}"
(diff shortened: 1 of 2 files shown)


checks/normandy/remotesettings_recipes.py
```diff
@@ -1,35 +1,52 @@
...
 
+    # Make sure the baseline recipes are all listed in the baseline collection
     missing = []
     for rid, r in normandy_by_id.items():
-        published = remotesettings_by_id.pop(rid, None)
+        published = rs_baseline_by_id.pop(rid, None)
         if published is None:
             missing.append({"id": r["id"], "name": r["name"]})
-    extras = [{"id": r["id"], "name": r["name"]} for r in remotesettings_by_id.values()]
+    extras = [{"id": r["id"], "name": r["name"]} for r in rs_baseline_by_id.values()]
```
glasserc: Should we check that the capabilities recipes are all listed in the capabilities collection too? Or is that somewhere else?

leplatrem: You're right. I first thought they were not available via the Normandy API, but passing  `only_baseline_capabilities=0` returns them. I'll update the code

PR #141, opened 2019-10-28
Fix #139: Support tags in addition to releases to compare deployed version

### LINKED ISSUE #139: checks.core.deployed_version doesn't work against mozilla/normandy
The reason is because https://api.github.com/repos/mozilla/normandy/releases returns `[]`, an empty array.  This is because the [documentation](https://developer.github.com/v3/repos/releases/) say: 

> This returns a list of releases, which does not include regular Git tags that have not been associated with a release. To get a list of Git tags, use the Repository Tags API.

I suppose this means several things: 

1. Standardize on the way we do releases ...
2. Change the check to pull tags instead of releases
3. ??

leplatrem:
Fix #139

### DIFF
--- a/checks/core/deployed_version.py
+++ b/checks/core/deployed_version.py
@@ -12,12 +12,16 @@ EXPOSED_PARAMETERS = ["server", "repo"]
 async def run(server: str, repo: str) -> CheckResult:
     # Server __version__ (see mozilla-services/Dockerflow)
     version_info = await fetch_json(server + "/__version__")
-
-    # Latest Github tag.
-    releases = await fetch_json(f"https://api.github.com/repos/{repo}/releases")
-
     deployed_version = version_info["version"]
-    latest_tag = releases[0]["tag_name"]
+
+    # Latest Github release.
+    releases = await fetch_json(f"https://api.github.com/repos/{repo}/releases")
+    if len(releases) > 0:
+        latest_tag = releases[0]["tag_name"]
+    else:
+        # Let's use the tag list if no release is defined.
+        tags = await fetch_json(f"https://api.github.com/repos/{repo}/tags")
+        latest_tag = tags[0]["name"]
 
     return (
         deployed_version == latest_tag,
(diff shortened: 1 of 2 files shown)


leplatrem:
This does not really work actually, since tags are returned as sorted alphabetically and order cannot be changed. Apparently the GraphQL endpoint allows to sort by the datetime of the associated commit.

leplatrem:
I modified the code to obtain the tags order by commit, but I am not sure anymore if this makes sense :) This introduces complexity and the GraphQL API requires a Github auth token. Using the *releases* feature for Normandy would be easier :wink:

leplatrem:
> I think it is isn't valid to only look at the first page of this API.

I rewrote the tags query using GraphQL (it returns the last entry)

checks/core/deployed_version.py
```diff
@@ -4,20 +4,57 @@
...
+              repository(owner: "%s", name: "%s") {
+                refs(refPrefix: "refs/tags/", last: 1, orderBy: {field: TAG_COMMIT_DATE, direction: ASC}) {
+                  edges {
+                    node { name }
+                  }
+                }
+              }
+            }
+        """
+        % (owner, name)
```
glasserc: Do we have to quote-escape these or something? I guess it's probably fine because we expect these values to come from configuration..

leplatrem: Yeah, I wouldn't mind if poucave fails with quotes in owner/project strings :)

leplatrem:
Do you think we should make the Github token parameter optional? If the project uses the releases feature of Github, then we don't need to fetch stuff from the auth-protected GraphQL API

glasserc:
To be honest, I already find the blurriness of the reasoning ("if there were ever releases, don't check tags") a little scary. It means changes in workflow kind of break the project. I would generally prefer that we always use tags, except that there's no guarantee that all tags are releases either. (You might create a tag to mark some last-known-good version, or a maintenance release, or something else..) I'm not convinced that we can do anything that is 100% correct here, so I would suggest we stick with what's simple.

leplatrem:
> so I would suggest we stick with what's simple.

Then only Github releases... No bearer token, no confusion with non-release tags etc.

PR #160, opened 2019-11-05
Fix #106: Leverage the autograph-utils lib

### LINKED ISSUE #106: How do we verify certificate chains?
I am comparing the Remote Settings cert verification chain code against Normandy's to try to resolve https://github.com/mozilla/normandy/issues/1890. The Normandy code is quite a bit more complicated than the `x5u`-checking code we have in this repository. In particular, it looks like the Normandy code parses a [series of certificates](https://github.com/mozilla/normandy/blob/master/normandy/recipes/signing.py#L198). I'm not sure how the `cryptography` library works, but it seems like we [only get one cert](https://github.com/mozilla-services/poucave/blob/653c0a6536d2b01ae9b0847b95e1c332b89b6e23/checks/remotesettings/validate_signatures.py#L53). What happens to the rest of the chain?

leplatrem:
Fix #106

* [x] Wait for offcial release
* [x] Fix tests coverage
* [ ] Add tests for specific exceptions?

### DIFF
--- a/checks/remotesettings/certificates_expiration.py
+++ b/checks/remotesettings/certificates_expiration.py
@@ -7,13 +7,24 @@ expiration date and x5u URL.
 import datetime
 import logging
 
+import cryptography
+import cryptography.x509
+from cryptography.hazmat.backends import default_backend as crypto_default_backend
+
 from poucave.typings import CheckResult
-from poucave.utils import run_parallel, utcnow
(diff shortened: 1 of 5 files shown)


leplatrem:
Blocked on https://github.com/mozilla-services/python-autograph-utils/pull/8

PR #164, opened 2019-11-05
Fix #149: Add endpoint for all checks by project

### LINKED ISSUE #149: Add endpoint to return result of all checks
We could add an endpoint `/checks/{project}`  that would return the aggregation of all results of a project.

It would use the cached values if any available

```
[{
  "name": "preview-signatures",
  "project": "remotesettings",
  "module": "checks.remotesettings.validate_signatures",
  "description": "Signatures validation of preview collections",
  "documentation": "Signatures should be valid for each collection content.\n\nThe errors are returned for each concerned collection.",
  "url": "/checks/remotesettings/preview-signatures",
  "ttl": 3600,
  "parameters": {},
  "datetime": "2019-10-30T15:12:09.266873",
  "success": true,
  "data": {}
}, {
...
}]
```

leplatrem:
Fix #149

### DIFF
--- a/poucave/app.py
+++ b/poucave/app.py
@@ -18,6 +18,8 @@ from . import config, middleware, utils
 
 HTML_DIR = os.path.join(os.path.dirname(os.path.realpath(__file__)), "html")
 
+routes = web.RouteTableDef()
+
 
 class Checks:
     def __init__(self, conf):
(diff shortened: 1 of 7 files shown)


leplatrem:
It's based on #165. Review by commit shouldn't be too painful :pray:

poucave/app.py
```diff
@@ -19,6 +19,32 @@
...
 
+class Checks:
+    def __init__(self, conf):
+        self.all = []
+        for project, checks in conf["checks"].items():
+            for name, params in checks.items():
+                check = Check(project, name, **params)
+                self.all.append(check)
+
+    def get(self, project=None, name=None):
```
glasserc: You proposed `lookup` in a video call, and I think that makes more sense (it fails if missing). But I think we should define another method for `get_all`.

poucave/app.py
```diff
@@ -19,6 +19,32 @@
 HTML_DIR = os.path.join(os.path.dirname(os.path.realpath(__file__)), "html")
 
 
+class Checks:
+    def __init__(self, conf):
+        self.all = []
+        for project, checks in conf["checks"].items():
+            for name, params in checks.items():
+                check = Check(project, name, **params)
+                self.all.append(check)
```
glasserc: Parsing out from `conf` seems a little odd, can we maybe make this a `classmethod` called `from_conf` or something?

poucave/app.py
```diff
@@ -144,6 +152,43 @@ def override_params(self, params: Dict[str, Any]):
...
+    cache = request.app["poucave.cache"]
+
+    try:
+        selected = checks.get(**request.match_info)
+    except ValueError:
+        raise web.HTTPNotFound()
+
+    body = []
+    for check in selected:
+        result = cache.get(check.cache_key)
```
glasserc: Should we be handling overridden parameters from URLs?

poucave/app.py
```diff
@@ -144,6 +152,43 @@ def override_params(self, params: Dict[str, Any]):
...
+        if result is None:
+            # This check never ran successfully.
+            body.append(check.infos)
+            continue
```
glasserc: I feel like we should be executing checks as normal here.

poucave/app.py
leplatrem: I don't think so... that would mean the parameters with the same names would have the same value on every check?

PR #176, opened 2019-11-07
Fix #170: cached timestamps carry timezone info

### LINKED ISSUE #170: In UI, timeago compares local date with UTC
Which leads to weird values like `In 7 hours` :)

leplatrem:
Fix #170

This should be enough for timeago to show relative time properly

### DIFF
--- a/tests/test_basic_endpoints.py
+++ b/tests/test_basic_endpoints.py
@@ -1,3 +1,4 @@
+import re
 from unittest import mock
 
 from poucave import config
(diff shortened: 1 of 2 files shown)


tests/test_basic_endpoints.py
```diff
@@ -79,7 +80,7 @@
 
     assert response.status == 200
     body = await response.json()
-    assert "datetime" in body
+    assert re.compile("....-..-..T..:..:..\\.......+..:..").match(body["datetime"])
```
mostlygeek: can't we just make this numbers instead of wildcard?

PR #177, opened 2019-11-07
Use UTC timezone in all datetimes

leplatrem:


glasserc:
Just for the sake of completeness, the timezone recommendations we follow in this project are https://julien.danjou.info/python-and-timezones/, right?

leplatrem:
> Just for the sake of completeness, the timezone recommendations we follow in this project are https://julien.danjou.info/python-and-timezones/, right?

From what I can read, yes

PR #203, opened 2019-11-15
Fix #187: swallow CancelledErrors

### LINKED ISSUE #187: Loads of CancelledError poucave.main.Handlers.checkpoint.<locals>.handler

leplatrem:
Fix #187 

They don't really bring any value when logged in Sentry. Let's swallow them and return a 500

### DIFF
--- a/poucave/middleware.py
+++ b/poucave/middleware.py
@@ -1,3 +1,4 @@
+import asyncio
 import logging
 import time
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


glasserc:
From video call: these come up when the client (the browser) disconnects from the server (Poucave). This happens when we have a request that takes too long -- `nginx` breaks the connection and we get this exception. Surfacing the exception as part of the request seems reasonable but it doesn't seem helpful for them to go to Sentry.

PR #225, opened 2019-11-28
Fix #222: Add check for Normandy JEXL error rate

### LINKED ISSUE #222: Add a check for normandy-client failing to communicate with classify client
We have [bugs like this](https://bugzilla.mozilla.org/show_bug.cgi?id=1594954) where we see mass unenrollments when the classify-client service is unavailable.  This is bad and messes up experiment date.   It's seems to be caused by the decision:  ["filter expressions will gracefully fail"](https://searchfox.org/mozilla-central/source/toolkit/components/normandy/lib/RecipeRunner.jsm#284-291). 

There will be a longer term fix.  Likely resulting in a chance in behavior on the client side.  

Poucave should monitor for event telemetry for errors talking to classify-client.  When the error rate exceeds a threshold the check should fail.

leplatrem:
Fix #222

### DIFF
--- a/checks/normandy/classify_error_rate.py
+++ b/checks/normandy/classify_error_rate.py
@@ -1,0 +1,41 @@
+"""
+The percentage of classify client errors in Normandy should be under the specified
+maximum.
+
+The error rate percentage is returned The min/max timestamps give the datetime range of the
+dataset obtained from https://sql.telemetry.mozilla.org/queries/64921/
+"""
+from poucave.typings import CheckResult
+from poucave.utils import fetch_redash
+
+
+EXPOSED_PARAMETERS = ["max_error_percentage"]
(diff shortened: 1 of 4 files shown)


checks/normandy/classify_error_rate.py
```diff
@@ -0,0 +1,41 @@
+"""
+The percentage of classify client errors in Normandy should be under the specified
+maximum.
+
+The error rate percentage is returned The min/max timestamps give the datetime range of the
```
mythmon: Something seems wrong here. Is there a word missing or something?

leplatrem:
> That being said, this is still a thing we should definitely be monitoring. I think tweaking the names/comments would be enough.

:+1:

PR #241, opened 2019-12-09
 Fix #214: run checks in background 

### LINKED ISSUE #214: Run checks in the background
The aiohttp server toolkit allows to have background tasks running.  https://docs.aiohttp.org/en/stable/web_advanced.html#background-tasks 

I was thinking that maybe we could have one that iterates through all the checks and executes them periodically?

leplatrem:


leplatrem:
I'm thinking that even if this relatively straightforward to implement (as this PR shows), we may follow the path of using an external ping service instead. Especially now that we have #164

PR #243, opened 2019-12-10
Fix #240: add check to verify Normandy recipes signatures

### LINKED ISSUE #240: Verify Normandy signatures in published remote settings collections
In Normandy, we recently made [a change](https://github.com/mozilla/normandy/pull/2053) that resulted in the signatures for recipes being incorrect in the `normandy-recipes-capabilities` collection only. The recipe data published to this collection was different than the recipe data that Normandy had gotten signed. This caused signatures to fail on all clients. Luckily QA found this before it reached production.

To find this problem quicker in the future, we should verify the Normandy signatures on the collections `normandy-recipes` and `normandy-recipes-capabilities`. Note that these signatures are separate from the Remote Settings signature of the collection.

leplatrem:
Fix #240

(I know that Ethan has a lot of stuff in his review queue ;))


checks/normandy/recipes_signatures.py
```diff
@@ -0,0 +1,47 @@
...
+async def validate_signature(verifier, recipe):
+    signature_payload = recipe["signature"]
+    x5u = signature_payload["x5u"]
+    signature = signature_payload["signature"]
+    attributes = recipe["recipe"]
+    data = json.dumps(attributes, sort_keys=True, separators=(",", ":")).encode("utf8")
+    return await verifier.verify(data, signature, x5u)
+
+
+async def run(server: str, collection: str, root_hash: str) -> CheckResult:
```
mythmon: It would be nice to have a doc string explaining the expected values for these parameters, or to include them in the module doc string.

checks/normandy/recipes_signatures.py
```diff
@@ -0,0 +1,47 @@
+"""
```
mythmon: I think the filename of this should be `recipe_signatures.py`, instead of `recipes_signatures.py`. That is, only "signatures" is plural, since English is weird about plurals and adjectives.

leplatrem:
Thanks for the feedback!

PR #244, opened 2019-12-10
Adjust consistency check description to reflect software issue likeliness

leplatrem:
This will be shown in the Web UI


mostlygeek:
Are we going to link to mana?

leplatrem:
> Are we going to link to mana? That way we can have a page to capture notes and fixes for posterity.
Merge state

That would be a different issue/PR

PR #247, opened 2019-12-11
Add link to troubleshooting docs

leplatrem:
@mostlygeek suggested one dedicated page per check. I think we can start with one section first, and if it grows too much then split.

What do you think Ethan?

### DIFF
--- a/poucave/app.py
+++ b/poucave/app.py
@@ -128,6 +128,9 @@ class Check:
 
     @property
     def info(self):
+        troubleshooting_url = config.TROUBLESHOOTING_LINK_TEMPLATE.format(
+            project=self.project, check=self.name
+        )
         return {
             "name": self.name,
             "project": self.project,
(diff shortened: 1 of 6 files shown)


mostlygeek:
This doesn't match what I envisioned.  I'd like to provide more clarity on what I want and how I'm thinking about this. 

The problem: 

Fixing a failing check requires knowledge and experience.  Those two things exist in only a few people's brains.  People suck as knowledge archives.   A wiki is much better.  

My idea for using a wiki: 

- link to mana.  wikis create pages dynamically so no need to create pages beforehand. 
- one page per check.  Some checks are for complex behavior.  Solutions may also be complex.  A page per check has a better signal to noise ratio. 
- anyone (moz employee) can read and write to the page
- I assume people are going to mana in an anxious state and looking for solutions
- page should be organized like so: 
  - at the top: How to fix the failing check
  - Information and context on the check, stuff that doesn't belong in GH issues
  - Archive of previous break/fix notes and how the problem was solved
  - ???

leplatrem:
Thanks for the feedback!

> link to mana. wikis create pages dynamically so no need to create pages beforehand.

I was looking at this, and I didn't find how to do it (Confluence doesn't seem to work like Mediawiki). There is a nice Troubleshooting template though, I could only trigger from the menu...

PR #294, opened 2020-01-09
Add specific header to each request

leplatrem:


tests/checks/remotesettings/test_utils.py
```diff
@@ -112,3 +113,30 @@ def test_kinto_auth():
 
     assert client._client.session.auth.type == "Bearer"
     assert client._client.session.auth.token == "token"
+
+
+async def test_client_extra_headers(mock_responses):
+    server_url = "http://fake.local/v1"
+    mock_responses.get(server_url + "/", payload={})
+    backup = config.DEFAULT_REQUEST_HEADERS
+    config.DEFAULT_REQUEST_HEADERS = {"Extra": "header"}
```
smarnach: How about decorating this function with

    @unittest.mock.patch("config.DEFAULT_REQUEST_HEADERS", {"Extra": "header"})

Monkey patching is never great for test isolation, but this at least makes sure it's reset in case the test fails.

PR #311, opened 2020-01-21
Show min/max error rate in uptake checks

leplatrem:


checks/remotesettings/uptake_error_rate.py
```diff
@@ -140,6 +147,8 @@ def sort_dict_desc(d, key):
         },
         ...
       },
+      "min_rate": 2.1,
+      "max_rate": 4.2,
```
glasserc: Should this be `6.12` to correspond to the example?

PR #312, opened 2020-01-21
Ignore uptake status on a particular version

leplatrem:
This will allow us to ignore `parse_error`s on 68. 

Should we specify the collection too? 
I was wondering what could be the syntax so I gave up :smirk: ...
```toml
ignore_status = ["security-state/intermediates:parse_error@68"]
```


glasserc:
Why do we have an excess of `parse_error` on 68?

leplatrem:
> Why do we have an excess of parse_error on 68?

The fix was not backported https://bugzilla.mozilla.org/show_bug.cgi?id=1558812#c12

PR #325, opened 2020-02-03
Fix #324: add ability to refresh cache 

### LINKED ISSUE #324: Add ability to refresh check result
https://bugzilla.mozilla.org/show_bug.cgi?id=1612688

There should be a way to force the refresh of a check result.

Since some checks pull a lot of info from the related services, process a lot of data, and run with a single worker (cache is in memory), then we should also provide some protection against DDOS

leplatrem:


leplatrem:
@smarnach r?

poucave/app.py
```diff
@@ -264,13 +264,18 @@ def override_params(self, params: Dict[str, Any]):
     except ValueError:
         raise web.HTTPNotFound()
 
+    # Refresh cache?
+    force = "refresh" in request.query
+    if force and request.query["refresh"] != config.REFRESH_SECRET:
+        raise web.HTTPBadRequest()
```
smarnach: Including an error message here would be nice.

README.md
```diff
@@ -42,6 +42,8 @@ Other endpoints:
 * ``/checks/{a-project}``: execute all checks of project ``a-project``
 * ``/checks/tags/{a-tag}``: execute all checks with tag ``a-tag``
 
+Cache can be forced to be bypassed with ``?refresh={s3cr3t}``. See *Environment variables* section.
```
smarnach: This makes it sound like the parameter works for the "other endpoints" above, while the code looks like it only works for individual checks (not for projects or tags). Could we clarify this?

PR #328, opened 2020-02-05
Render JSON when Accept:*/* is specified

leplatrem:


poucave/utils.py
```diff
@@ -148,7 +148,14 @@ def render_checks(func):
     async def wrapper(request):
         # First, check that client requests supported output format.
         is_text_output = False
-        accepts = set(request.headers.getall("Accept", []))
+        # Several Accept headers can be specified, and each of them can have
+        # comma-separated values (eg. `Accept: text/html,application/xhtml+xml`).
```
smarnach: The full format is [rather complex](https://tools.ietf.org/html/rfc7231#section-5.3.2). Can we use a library to parse the header? While we don't have to support all formats clients could possibly send, using a lib would reduce the likelihood we have to fix more bugs in this code – it's starting to become non-trivial already.

I know that there is `parse_accept_header` in Werkzeug, which is what I used to use when working with Flask. What do you think?

poucave/utils.py
```diff
@@ -148,7 +148,14 @@ def render_checks(func):
...
+        accepts = set(
+            chain(*[a.split(",") for a in request.headers.getall("Accept", [])])
```
smarnach: There's also `chain.from_iterable()`, which is meant for this use case. (A pity they gave it that stupid name instead of just calling it `flatten()`.)

poucave/utils.py
```diff
@@ -148,7 +148,14 @@ def render_checks(func):
...
+        )
+        # Also some values can specify a "quantity" value that we can ignore here (eg. `*/*;q=0.8`)
+        accepts = {a.split(";")[0] for a in accepts}
```
smarnach: If we decide keeping this code, I suggest adding `.strip()` here, since spaces after the commas are allowed.

leplatrem:
I agree, we should either rely on some library or keep it super simple. I'd vote for the later, ie. a simple `if "text/plain" in ...`.

PR #329, opened 2020-02-06
Improve reported recipes check

leplatrem:
This will allow us to have several checks by channel, with different thresholds. For example:

```toml

[checks.normandy-uptake.reported-recipes-release]
description = "Normandy reported recipes on ESR/Release"
module = "checks.normandy.reported_recipes"
params.api_key = "XXXXXXXXX"
params.server = "https://normandy.cdn.mozilla.net"
params.channels = ["release", "esr"]
params.min_total_events = 500  # Sampled 1% means 50K

[checks.normandy-uptake.reported-recipes-beta]
description = "Normandy reported recipes on Beta/Aurora"
module = "checks.normandy.reported_recipes"
params.api_key = "XXXXXXXXX"
params.server = "https://normandy.cdn.mozilla.net"
params.channels = ["beta", "aurora"]
params.min_total_events = 1000

[checks.normandy-uptake.reported-recipes-nightly]
description = "Normandy reported recipes on Nightly"
module = "checks.normandy.reported_recipes"
params.api_key = "XXXXXXXXX"
params.server = "https://normandy.cdn.mozilla.net"
params.channels = ["nightly"]
params.min_total_events = 100
```

where the API key can be obtained from https://sql.telemetry.mozilla.org/queries/67658/


checks/normandy/reported_recipes.py
```diff
@@ -29,7 +29,11 @@
 
 
 async def run(
-    api_key: str, server: str, min_total_events: int = 1000, lag_margin: int = 600
+    api_key: str,
+    server: str,
+    min_total_events: int = 1000,
+    lag_margin: int = 600,
+    channels: List[str] = [],
```
mythmon: In Rust I'd suggest the type of this be `Option<Vec>`, instead of giving special meaning to the empty list. Do you think that's worth doing here? (ie, `Optional[List[str]]`) ([docs on Optional](https://docs.python.org/3/library/typing.html#typing.Optional))

leplatrem:
> As a side note, this is starting to make me think of some code I have in Edi where I wanted an async parallel pipeline, and I found the library pypeln very useful. I wonder if Poucave could benefit from it as well?

Indeed, this library looks neat! We were doing everything with [these couple of lines](https://github.com/mozilla-services/poucave/blob/7c758971e523f7baaa726c8e2a053e5f31a1b8ea/poucave/utils.py#L88-L135), but I'd be happy to replace them with a more powerful library :) Thanks for the suggestion!

PR #330, opened 2020-02-07
Add status parameter to heartbeat check

leplatrem:
This would allow us to have a check for the CDN which returns 403 on its root URL

### DIFF
--- a/checks/core/heartbeat.py
+++ b/checks/core/heartbeat.py
@@ -9,15 +9,15 @@ from poucave.typings import CheckResult
 from poucave.utils import ClientSession, retry_decorator
 
 
-EXPOSED_PARAMETERS = ["url", "status"]
+EXPOSED_PARAMETERS = ["url", "expected_status"]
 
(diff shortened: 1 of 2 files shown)


checks/core/heartbeat.py
```diff
@@ -9,15 +9,15 @@
...
 
 
 @retry_decorator
-async def run(url: str) -> CheckResult:
+async def run(url: str, status: int = 200) -> CheckResult:
     async with ClientSession() as session:
         try:
             async with session.get(url) as response:
-                status = response.status == 200
+                status = response.status == status
```
smarnach: You are shadowing the integer parameter `status` with a Boolean variable of the same name here, which makes this slightly confusing on first reading. How about renaming the new parameter to `expected_status` instead? This might be a better description of what it does anyway, and make the configuration more readable as well.

leplatrem: You're fully right :)

PR #333, opened 2020-02-07
Fix #332: Add check for Normandy runner uptake

### LINKED ISSUE #332: Add check for Normandy runner uptake
Currently the Redash query filters events with ```source LIKE 'normandy/recipe/%'```, and the runner uptake status is not «monitored».

leplatrem:
Fix #332


leplatrem:
There should be a better approach, reusing the same Redash source as the one of recipes

PR #339, opened 2020-02-10
Fix #331: Add SVG diagram feature

### LINKED ISSUE #331: Add SVG diagram to HTML UI
Poucave would take an optional config `DIAGRAM_PATH=./diagram.svg` to be shown in the HTML UI.

Elements of the SVG would have specific IDs (eg. `#check-remotesettings/writer-heartbeat`, `#check-{project}/{name}`), that would be toggled red or green in JavaScript.

```html
    <object id="svg-diagram" data="path/to/diagram.svg" type="image/svg+xml"></object>
```

```js
    const svgObject = document.getElementById('svg-diagram').contentDocument;
    const element = svgObject.getElementById('check-remotesettings/writer-heartbeat');
    element.setAttribute( ... );
```

This way we could have a whole picture of the pipeline architecture, with real time status :)

leplatrem:
* [x] Play with API (fill for shapes / stroke for lines etc.)
* [x] Clean code
* [x] Read SVG from current path (instead of repo)
* [x] Add tooltips


poucave/html/index.html
```diff
@@ -5,13 +5,16 @@
   <link rel="icon" href="" type="image/png" sizes="32x32"></link>
 </head>
 <body>
+  <object id="svg-diagram" data="../diagram.svg" type="image/svg+xml"></object>
```
mostlygeek: * just make this `/diagram.svg`?, it's served from the root anyways.  No need for a relative URL 
* Having this `<object>` says:  a diagram MUST exist.   A diagram should be optional.

poucave/html/style.css
```diff
@@ -29,10 +29,15 @@ body {
   color: var(--text-color);
 }
 
-a, a:hover, a:visited {
+a[href^="+"], a:hover[href^="+"], a:visited[href^="+"] {
   color: var(--link-color);
 }
 
+#svg-diagram {
```
mostlygeek: I don't think we should enforce a `max-height`.   By default if the SVG 404's ... will the height be 0px?

poucave/html/script.js
```diff
@@ -25,10 +25,24 @@ async function refreshCheck(check, options = {}) {
     }
   }
 
+  // In SVG diagram too :) 
+  const svgObject = document.getElementById('svg-diagram').contentDocument;
+  const svgElement = svgObject ? svgObject.getElementById(`${check.project}--${check.name}`) : null;
```
mostlygeek: A string made up of "{project}--{name}" is brittle.  It is not obvious this relationship between project/check name and values baked into an SVG.  Maybe a warning in the documentation is enough to save someone (us!) some debugging time in the future.

poucave/html/index.html
leplatrem: > Having this <object> says: a diagram MUST exist. A diagram should be optional.

I don't agree. It will just show a 404. Like a favicon or whatever.

poucave/html/style.css
leplatrem: Yes height will be 0px.

I thought the max-height would make sense with `width:100%`. It basically tries to fit horizontally without taking too much vertical space. I changed to 33% of vertical screen instead of static 400px

poucave/html/index.html
mostlygeek: favicons should not 404 either. For this it probably doesn't matter as the traffic and impact will be quite small.

mostlygeek:
I have a different opinion on the 404s for missing svgs but it doesn't need to block this

leplatrem:
I like the explicit and loud 404 ;) 

Merging :)

leplatrem:
Screenshot_2020-02-13 Screenshot

PR #340, opened 2020-02-11
Filter Recipes uptake based on list of enabled recipes

leplatrem:
This change will require the uptake checks to receive an additional `server` parameter (even though for Telemetry we can only look at prod, since we have no way to distinguish stage recipes from prod)

### DIFF
--- a/checks/normandy/uptake_error_rate.py
+++ b/checks/normandy/uptake_error_rate.py
@@ -60,7 +60,7 @@ async def run(
     # Fetch list of enabled recipes from Normandy server.
     normandy_url = NORMANDY_URL.format(server=server)
     normandy_recipes = await fetch_json(normandy_url)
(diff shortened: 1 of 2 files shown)


checks/normandy/uptake_error_rate.py
```diff
@@ -11,13 +11,15 @@
...
 from poucave.typings import CheckResult
-from poucave.utils import fetch_redash
+from poucave.utils import fetch_json, fetch_redash
 
 
 EXPOSED_PARAMETERS = ["max_error_percentage", "min_total_events"]
 
 REDASH_QUERY_ID = 67658
 
+NORMANDY_URL = "{server}/api/v1/recipe/signed/?enabled=1"
```
mythmon: At some point we need to get Poucave to use the v3 API or GraphQL. I'd like to avoid using v1 much longer. That being said, this kind of "get all enabled recipes" isn't super easy in the v3 API, so I'm not going to insist you change it here.

PR #341, opened 2020-02-11
Compute JEXL error rate by period

leplatrem:


checks/normandy/jexl_error_rate.py
```diff
@@ -21,13 +24,29 @@
...
     max_timestamp = max(r["max_timestamp"] for r in rows)
 
-    total = sum(row["total"] for row in rows)
-    classify_errors = sum(
-        row["total"] for row in rows if row["status"] == "content_error"
-    )
-    error_rate = classify_errors * 100.0 / total
+    # The Redash query returns statuses by periods (eg. 10min).
+    # First, agregate totals by period.
+    periods: Dict[Tuple[str, str], Dict] = {}
```
mythmon: I think this would work nicer as a defaultdict with a `Counter` as the default value

```python
from collections import Counter
periods = defaultdict(Counter)
```

I'm not sure how to do the typing on that though

checks/normandy/jexl_error_rate.py
```diff
@@ -21,13 +24,29 @@
...
+
+    # Then, keep the period with highest error rate.
+    max_error_rate = 0.0
+    for period, all_statuses in periods.items():
+        total = sum(all_statuses.values())
+        classify_errors = all_statuses.get("content_error", 0)
+        error_rate = classify_errors * 100.0 / total
+        if max_error_rate < error_rate:
+            max_error_rate = error_rate
+        if max_error_rate >= max_error_percentage:
```
mythmon: Should this be `>` instead of `>=`, since it pairs with `<=` in the return statement?

checks/normandy/jexl_error_rate.py
```diff
@@ -21,13 +24,29 @@
...
+        periods[period][status] += row["total"]
+
+    # Then, keep the period with highest error rate.
+    max_error_rate = 0.0
+    for period, all_statuses in periods.items():
+        total = sum(all_statuses.values())
+        classify_errors = all_statuses.get("content_error", 0)
+        error_rate = classify_errors * 100.0 / total
+        if max_error_rate < error_rate:
+            max_error_rate = error_rate
```
mythmon: Instead of the if block, what about this?

```python
max_error_rate = max(max_error_rate, error_rate)
```

checks/normandy/jexl_error_rate.py
leplatrem: Awesome :) TIL about `Counter`!

checks/normandy/jexl_error_rate.py
leplatrem: Aww.. of course :roll_eyes:

checks/normandy/jexl_error_rate.py
leplatrem: Good catch, thanks!

PR #347, opened 2020-02-17
Complete makeover of the UI

rehandalal:
I maybe went a little overboard this weekend. But I ended up doing a total refactor of the UI.

Overview of changes:
- Rewrote `script.js` to feel a little more like an SPA without needing a full framework.
- Projects and checks are now sorted by name.
- Added in a few new libraries: Tabler, FontAwesome.
- Redesigned the favicons.
- Redesigned the diagram SVG. (New one is [here](https://gist.github.com/rehandalal/5540af8f40350b4ceed879dfc4c6e3f7))

Still needs some work:
- ~The result details in each check are still messy and I haven't figured out the best way to present those yet.~
- The refresh button feels clunky to me.


rehandalal:
Some screenshots:

image

image

rehandalal:
> Maybe we should consider introducing a lightweight framework like Vue or Mithril? What do you think?

I was also thinking about this and was going to explore  [Preact](https://preactjs.com). It's a super lightweight (3kb) framework heavily based off of React that you can use right in the browser without transpiling. So no complicated build step.

.gitignore
```diff
@@ -1,8 +1,8 @@
 /config.toml
-/diagram.svg
 /version.json
 __pycache__
 /.venv
 *.pyc
 .coverage
 .mypy_cache
+poucave/html/img/diagram.svg
```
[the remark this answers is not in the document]
rehandalal: ah! yeah i had trouble getting the diagram to show up (I kept putting it in the root of `/html/`) and then this just worked so i ran with it...

i'll move it to the root of the project

leplatrem:
> I was also thinking about this and was going to explore Preact.

Yeah I heard good things :) I think it'd be worth it

rehandalal:
phew! so this is rewritten in Preact. I think it's a lot cleaner. it could still probably use a little refining and the interaction with the SVG diagram is still a little janky.

rehandalal:
ezgif com-gif-maker

Automatic light/dark mode included!

poucave/html/js/app.js
```diff
@@ -0,0 +1,627 @@
...
+      ${this.renderProjects()}
+    `;
+  }
+}
+
+class Overview extends Component {
+  render({ checks, results }) {
+    let isHealthy = true;
+
+    Object.keys(results).forEach(k => {
```
leplatrem: `const isHealthy = Object.values(results).every(r => r.success)`

poucave/html/js/app.js
```diff
@@ -0,0 +1,627 @@
...
+    const { diagramReady } = this.state;
+    const {checks, results} = this.props;
+
+    const svgDoc = diagramReady ? this.svgRef.current.contentDocument : null;
+
+    Object.keys(results).forEach(k => {
+      const c = checks[k];
+      const r = results[k];
+
+      if (svgDoc) {
```
leplatrem: What do you think of a  `if (!svgDoc) { return; }`  above?

poucave/html/js/app.js
```diff
@@ -0,0 +1,627 @@
...
+            });
+          }
+
+          indicator.removeAttribute("fill");
+          indicator.classList.remove("fill-gray", "fill-red", "fill-green");
+          let fillClass = "fill-gray"
+          if (!r.isLoading) {
+            fillClass = r.success ? "fill-green" : "fill-red";
+          }
+          indicator.classList.add(fillClass);
```
leplatrem: Could we avoid duplication with initialization block above?

poucave/html/js/app.js
```diff
@@ -0,0 +1,627 @@
...
+    const { diagramHidden, diagramReady } = this.state;
+    const { results } = this.props;
+    const diagramClass = diagramReady ? "" : "invisible";
+    const diagramCardClass = diagramHidden ? "d-none" : "";
+
+    let isLoading = false;
+    Object.keys(results).forEach(k => {
+      const r = results[k];
+      isLoading |= r.isLoading;
+    });
```
leplatrem: `const isLoading = Object.values(results).some(r => r.loading)` ?

poucave/html/js/app.js
```diff
@@ -0,0 +1,627 @@
...
+      <section class="project mt-3 pt-4">
+        <div class="float-right mt-1 lh-1">
+          ${this.renderStatus()}
+        </div>
+        <h3 class="mb-4">
+          <i class="fa fa-layer-group mr-2"></i>
+          <span class="project-name">${name}</span>
+        </h3>
+        <div class="project-cards mb-1">
+          ${checks.map(c => html`<${Check} data="${c.data}" result="${c.result}" fetchResult="${fetchCheckResult}" />`)}
```
leplatrem: Maybe keep the name `fetchCheckResult` so that we can follow the reference more easily

poucave/html/js/modules.js
```diff
@@ -0,0 +1 @@
..._.ownerSVGElement,i&&!c?[i]:s?null:l.slice.call(_.childNodes),a,i||u,c),S(a,o)}var re=function(e){var n=te.get(this);return n||(n=new Map,te.set(this,n)),(n=ne(this,n.get(e)||(n.set(e,n=function(e){for(var n,t,o=1,r="",_="",u=[0],l=function(e){1===o&&(e||(r=r.replace(/^\s*\n\s*|\s*\n\s*$/g,"")))?u.push(0,e,r):3===o&&(e||r)?(u.push(3,e,r),o=2):2===o&&"..."===r&&e?u.push(4,e,0):2===o&&r&&!e?u.push(5,0,!0,r):o>=5&&((r||!e&&5===o)&&(u.push(o,0,r,t),o=6),e&&(u.push(o,e,0,t),o=6)),r=""},i=0;i<e.length;i++){i&&(1===o&&l(),l(i));for(var c=0;c<e[i].length;c++)n=e[i][c],1===o?"<"===n?(l(),u=[u],o=3):r+=n:4===o?"--"===r&&">"===n?(o=1,r=""):r=n+r[0]:_?n===_?_="":r+=n:'"'===n||"'"===n?_=n:">"===n?(l(),o=1):o&&("="===n?(o=5,t=r,r=""):"/"===n&&(o<5||">"===e[i][c+1])?(l(),3===o&&(u=u[0]),o=u,(u=u[0]).push(2,0,o),o=0):" "===n||"\t"===n||"\n"===n||"\r"===n?(l(),o=2):r+=n),3===o&&"!--"===r&&(o=4,u=u[0])}return l(),u}(e)),n),arguments,[])).length>1?n:n[0]}.bind(f);export{f as h,re as html,oe as render,h as Component,T as createContext,V as useState,I as useReducer,q as useEffect,B as useLayoutEffect,$ as useRef,j as useImperativeHandle,z as useMemo,G as useCallback,J as useContext,K as useDebugValue};
```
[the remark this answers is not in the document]
rehandalal: oh sorry i should rename this to `htm_preact.module.js` which I copied verbatim from here:
https://unpkg.com/htm/preact/standalone.module.js. because we want to import from it as a module we cannot use the concatenation that we are using for the other libraries.

poucave/html/js/app.js
```diff
@@ -0,0 +1,627 @@
...
+    };
+
+    return html`<div dangerouslySetInnerHTML="${innerHTML}" />`;
+  }
+}
+
+// Clear the application container and render the application
+const appContainer = document.getElementById("app");
+appContainer.innerHTML = "";
+render(html`<${App} />`, appContainer);
```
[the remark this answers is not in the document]
rehandalal: yep, good catch

poucave/html/js/app.js
```diff
@@ -0,0 +1,627 @@
...
+            </a>
+          </p>
+        </details>
+      </div>
+    `;
+  }
+
+  renderFooter() {
+    const { data, result } = this.props;
+    const updated = result.datetime ? new Date(result.datetime) : new Date();
```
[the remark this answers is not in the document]
rehandalal: Yes, we could also hide this altogether. My thinking here was if there's no datetime then it likely errored out in which case it was in some senses last updated now

PR #358, opened 2020-02-25
remove @import for libs.css which does not exist

mostlygeek:
libs.css doesn't exist so it just 404s.  This @import is unnecessary


poucave/html/css/main.css
```diff
@@ -1,4 +1,3 @@
-@import "libs.css";
 @import "main-dark.css";
```
leplatrem: Technically this as well since it's also listed in the HTML.

PR #360, opened 2020-02-25
Fix issue in Safari and add better mobile support

rehandalal:
r?


mostlygeek:
Did you test that this actually fixes the safari issues? :) if so R+

PR #361, opened 2020-02-26
Inject stylesheet into SVG

rehandalal:
This should allow the new styling to work with the old SVG as well.

r?

### DIFF
--- a/poucave/html/css/svg-diagram.css
+++ b/poucave/html/css/svg-diagram.css
@@ -1,0 +1,69 @@
+@import url(https://fonts.googleapis.com/css?family=Source+Sans+Pro);
+
+svg {
+    visibility: visible !important;
+}
+
+.fill-none {
+    fill: none;
+}
+
+.fill-solid {
+    fill: #303645;
+}
+
+.stroke-solid {
+    stroke: #303645;
+}
+
+.text-clear {
+    color: #fff;
+}
+
+.text-solid {
+    color: #303645;
+}
+
+.ghostly {
+    fill: #f8f8f9;
+    stroke: #acb0b8;
+}
+
+.fill-gray {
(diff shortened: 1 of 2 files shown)


poucave/html/css/svg-diagram.css
```diff
@@ -0,0 +1,69 @@
+@import url(https://fonts.googleapis.com/css?family=Source+Sans+Pro);
```
leplatrem: I'm not a fan of pulling stuff from Google TBH
Can we serve the woff files ourselves?

poucave/html/js/app.js
```diff
@@ -690,7 +712,12 @@ class Markdown extends Component {
 }
 
 // Initialize the app on page load
-window.addEventListener("load", () => {
+let diagramStylesheet;
+window.addEventListener("load", async () => {
+  // Load the stylesheet to be injected into the SVG
+  const diagramStylesheetResponse = await fetch("/html/css/svg-diagram.css");
+  diagramStylesheet = await diagramStylesheetResponse.text();
```
leplatrem: Can we avoid the global variable approach? :) 

For example, can we do something like `<style id="svg-style">...</style>` + `document.getElementById("svg-style").textContent`?

Or can we retrieve the content of a link BTW? `<link rel="stylesheet" >`? 

Or can we put the SVG styles just as a string? Maybe we don't strokes or dark variation for dots...  What do you think?

poucave/html/js/app.js
```diff
@@ -292,6 +309,8 @@ class SystemDiagram extends Component {
       return;
     }
 
+    this.initializeDiagram();
+
```
leplatrem: Why do we need it both on after load and update? (deserves a comment if relevant)

Issue #362, opened 2020-02-26
Make result panel more «readable»


leplatrem:
I guess the combination of several factors makes the results harder to read in the new UI:
- number of lines is now limited
- smaller font size
- less contrast (gray on gray)
- smaller width

I actually don't know which one exactly would have to be adjusted to improve it, but I think we can improve readability :) 

![Screenshot from 2020-02-26 18-24-09](https://user-images.githubusercontent.com/546692/75370361-3f2df700-58c5-11ea-877f-27e3a045bd81.png)
![Screenshot from 2020-02-26 18-23-34](https://user-images.githubusercontent.com/546692/75370366-3fc68d80-58c5-11ea-9ae8-66a03a6a6fdb.png)

leplatrem:
Fixed in #680

PR #367, opened 2020-03-02
Ref #362: enlarge results blocks

leplatrem:
Ref #362

If there has to be a max height, then it should be at least 30 lines :) 

Screenshot from 2020-03-02 11-05-47
Screenshot from 2020-03-02 11-05-29
Screenshot from 2020-03-02 11-03-50

### DIFF
--- a/poucave/html/css/main.css
+++ b/poucave/html/css/main.css
@@ -47,7 +47,6 @@ section.project:not(:first-child) {
 
 .check-details pre {
     background: rgba(48, 54, 69, 0.032);
-    max-height: 10rem;
     padding-bottom: 0;
     padding-top: 0;
 }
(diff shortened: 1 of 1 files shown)


rehandalal:
I'm not convinced this is the right solution. A couple things:
- I like the padding for me it creates better visual separation of what you're looking at. I think a better solution may be to keep the padding as it was and lose the third column in the grid so you get a much wider card and thus more data.
- At 30 lines there's no longer a point in having a max-height, we should just get rid of the max-height.

Thoughts?

leplatrem:
>  lose the third column in the grid so you get a much wider card and thus more data.

Why not, but could the columns be wider instead? 

Screenshot from 2020-03-03 11-55-39

Compared to https://leplatrem.github.io/poucave-old-ui/ 

Some feedback as a daily user:
- the result is THE main information of a failing card
- I spend all the time reading result panes, hopefully it will get better soon :) 
- the description, documentation, tags, and parameters are mainly useful for new comers. I personally don't read them anymore

PR #399, opened 2020-04-01
Remove inconsistent field from remotesettings-recipes check

leplatrem:
This field is redundant with other fields, and makes the result a bit confusing.

It was defined like this: `caps.inconsistent = rs.base - rs.caps`

But when the capabilities collection is in-sync with Normandy, then this inconsistent list always equals to the list of baseline extras.

Let's try to prove it :)

```
rs.caps = n.caps
rs.base = n.base + base.extras
```

Hence:
```
caps.inconsistent = rs.base - rs.caps
caps.inconsistent = rs.base - n.caps
caps.inconsistent = base.extras + n.caps - n.caps
caps.inconsistent = base.extras
```

And also in the case the capabilities collection is not in-sync with Normandy, and if we consider that all baseline recipes are also listed in capabilities on the Normandy server, then the inconsistent list equals to the list of missing capabilities.

```
n.base - n.caps = 0
rs.base = n.base + base.extras
```

```
caps.missing = n.caps - rs.caps
rs.caps = n.caps - caps.missing
```

Hence:
```
caps.inconsistent = rs.base - rs.caps
caps.inconsistent = rs.base - n.caps + caps.missing
caps.inconsistent = n.base + base.extras - n.caps + caps.missing
caps.inconsistent = base.extras + caps.missing
```

### DIFF
--- a/tests/checks/normandy/test_remotesettings_recipes.py
+++ b/tests/checks/normandy/test_remotesettings_recipes.py
@@ -66,11 +66,11 @@ async def test_positive(mock_aioresponses):
         payload=[NORMANDY_RECIPE],
     )
     mock_aioresponses.get(
-        REMOTESETTINGS_BASELINE_URL, payload={"data": [REMOTESETTINGS_RECIPE]}
-    )
-    mock_aioresponses.get(
         REMOTESETTINGS_CAPABILITIES_URL,
         payload={"data": [REMOTESETTINGS_RECIPE, REMOTESETTINGS_RECIPE_WITH_CAPS]},
+    )
+    mock_aioresponses.get(
+        REMOTESETTINGS_BASELINE_URL, payload={"data": [REMOTESETTINGS_RECIPE]}
     )
 
     status, data = await run(NORMANDY_SERVER, REMOTESETTINGS_SERVER)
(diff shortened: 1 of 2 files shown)


tests/checks/normandy/test_remotesettings_recipes.py
```diff
@@ -65,24 +65,25 @@
...
-    mock_aioresponses.get(
-        REMOTESETTINGS_BASELINE_URL, payload={"data": [REMOTESETTINGS_RECIPE]}
-    )
     mock_aioresponses.get(
         REMOTESETTINGS_CAPABILITIES_URL,
         payload={"data": [REMOTESETTINGS_RECIPE, REMOTESETTINGS_RECIPE_WITH_CAPS]},
     )
+    mock_aioresponses.get(
+        REMOTESETTINGS_BASELINE_URL, payload={"data": [REMOTESETTINGS_RECIPE]}
+    )
```
glasserc, line 74: Just to confirm, moving these calls around doesn't change anything, right? It was part of the same commit so I wanted to make sure it didn't have a semantic meaning.

leplatrem: No, I just wanted to have mocking in the same order as consuming

leplatrem:
> n.base - n.caps != 0 (for whatever reason)

It could belong to this check, but it basically consists in checking Normandy implementation, and I'm not sure it we want poucave to go that far

PR #402, opened 2020-04-02
Fix #380: Distinguish check failures

### LINKED ISSUE #380: Split sentry projects into one for code and one for checks
The `poucave-prod` project in sentry collects both code failures in the production instance of poucave as well as the checks that have failed, and using sentry to store the failed checks is a super clever way to get alerting and graphing for those checks and I love it, but perhaps it would be cleaner to separate those into two separate sentry projects so that we don't conflate one kind of failure with the other?

leplatrem:
Fix #380

From what I could read in the Sentry docs, we should be able to route our events based on tags (`source=check`).

@jaredlockhart r?

### DIFF
--- a/poucave/app.py
+++ b/poucave/app.py
@@ -97,9 +97,11 @@ class Check:
             self.params[param] = raw_type(value)
 
     async def run(self, cache=None, force=False) -> Tuple[Any, bool, Any, int]:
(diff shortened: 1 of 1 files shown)


leplatrem:
I'll take it as a r+ :)

PR #428, opened 2020-04-29
Fix #379: List failing checks on top

### LINKED ISSUE #379: Add links to failing checks to the  top of the page
Looking at the current landing page of Poucave, it strikes me that it would be helpful to add some more info here:

Screenshot 2020-03-06 17 12 19

that said something like

Failures: 2
Normandy: deployed-version
Normandy-Uptake-Beta: recipes-error-rate

that then link to each of the failing checks, so you could quickly see what's happening right at the top of the page.

leplatrem:
Fix #379

### DIFF
--- a/poucave/html/js/app.js
+++ b/poucave/html/js/app.js
@@ -142,6 +142,8 @@ class Dashboard extends Component {
         }
         console.warn(check.project, check.name, err);
         result = {
+          project: check.project,
+          name: check.name,
           success: false,
           datetime: new Date(),
           data: err.toString(),
@@ -224,7 +226,8 @@ class Dashboard extends Component {
 
 class Overview extends Component {
   render({ checks, results }) {
-    const isHealthy = Object.values(results).every(r => r.success);
+    const failing = Object.values(results).filter(r => !r.isLoading && !r.success);
+    const isHealthy = failing.length == 0;
 
     const iconClass = isHealthy ? "fa-check-circle text-green" : "fa-times-circle text-red";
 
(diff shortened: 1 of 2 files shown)


leplatrem:
Rehan could you please help make this look a bit better? 
Screenshot from 2020-04-29 17-29-57

poucave/html/js/app.js
```diff
@@ -246,12 +249,23 @@ class Overview extends Component {
           <div class="card-body text-center">
             <i class="fa fa-4x ${iconClass}"></i>
             <p>
-              <strong>The current system status is ${isHealthy ? "Healthy" : "Unhealthy"}.</strong>
+              <strong>The system ${isHealthy ? "is currently healthy" : "has failing checks"}.</strong>
```
rehandalal, line 252: 🎉

poucave/html/js/app.js
```diff
@@ -246,12 +249,23 @@ class Overview extends Component {
...
               <br />
               <span class="text-gray-medium">
                 Last updated <${TimeAgo} date="${new Date()}" />.
               </span>
             </p>
+            <p>
```
rehandalal, line 258: this should be a `ul` not a `p`. or there should be a `ul` inside this `p` tag to wrap the `li` nodes.

rehandalal: also i was probably give it a class of `text-red`

poucave/html/js/app.js
```diff
@@ -246,12 +249,23 @@ class Overview extends Component {
...
+              <${FocusedCheck.Consumer}>
+                ${focusedCheckContext => (
+                  failing.map(r => (
+                    html`<li>
+                      <a href="#" onClick=${() => focusedCheckContext.setValue(r.project, r.name)}>${r.project} / ${r.name}</a>
```
rehandalal, line 263: nit: wrap the content onto a new line:
```js
<a href="#" onClick=${() => focusedCheckContext.setValue(r.project, r.name)}>
  ${r.project} / ${r.name}
</a>
```

rehandalal: also i'd probably give this a class of `text-red` for better contrast

poucave/html/js/app.js
```diff
@@ -246,12 +249,23 @@ class Overview extends Component {
...
+                    </li>`
+                  ))
+                )}
+              </>
+            </p>
```
rehandalal, line 268: also it might be better to move this whole thing into a new method like:
```js
renderErrorList() {
  const failing = Object.values(results).filter(r => !r.isLoading && !r.success);
  if (failing.length > 0) {
    return "";
  }

  return html`
    <ul>
      ...
    </ul>
  `;
}
```

and then use here as:
```js
${this.renderErrorList()}
```

this will prevent an empty `<ul>`/`<p>` from being output.

PR #450, opened 2020-05-14
Avoid flickering of failing checks

leplatrem:
In order to keep of failing checks stable while reloading them, I made this trick. I wonder if there would be a better pattern, using the lifecycle methods for example.


leplatrem:
Hey Rehan, I'll merge and deploy without review. Once you give me your feedback I'll open another «refactor» PR if necessary

PR #451, opened 2020-05-14
Optimize latest-approvals requests

leplatrem:
This should greatly improve the check duration... 

See also https://bugzilla.mozilla.org/show_bug.cgi?id=1638044

### DIFF
--- a/checks/remotesettings/latest_approvals.py
+++ b/checks/remotesettings/latest_approvals.py
@@ -72,11 +72,20 @@ async def get_latest_approvals(
         previous = history[i + 1]
         after = previous["last_modified"]
         before = current["last_modified"]
+
+        # We are interested in the number of history entries between two approvals. The
+        # approval timestamps are part of the history object data (ie. `target` field)
+        # and are not indexed on the server. In order to reduce the cost of the request,
+        # we will prefilter the history entries by their own timestamp,
+        # assuming the history entries were created within 1sec after the target object
+        # modification (usually it's a few milliseconds).
         changes = await client.get_history(
             bucket=bucket,
             **{
                 "resource_name": "record",
                 "collection_id": collection,
+                "_since": after,
+                "_before": before + 1000,
                 "gt_target.data.last_modified": after,
                 "lt_target.data.last_modified": before,
             },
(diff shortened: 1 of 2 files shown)


checks/remotesettings/latest_approvals.py
```diff
@@ -72,11 +72,20 @@ async def get_latest_approvals(
         previous = history[i + 1]
         after = previous["last_modified"]
         before = current["last_modified"]
+
+        # We are interested in the number of history entries between two approvals. The
+        # approval timestamps are part of the history object data (ie. `target` field)
+        # and are not indexed on the server. In order to reduce the cost of the request,
+        # we will prefilter the history entries by their own timestamp,
+        # assuming the history entries were created within 1sec after the target object
+        # modification (usually it's a few milliseconds).
```
smarnach, line 81: Can you think of a query to verify the distribution of time differences between objects and history entries? We could also just deploy the change and see how it goes.

leplatrem: It would be something like that (*untested*)

```
SELECT age(last_modified, from_epoch(data->'target'->'data'->'last_modified'))
FROM objects
WHERE resource_name = 'history'
ORDER BY 1 DESC
LIMIT 100
``` 

Both timestamps are assigned in the same request/transaction when the change occurs, so 1000msec is already way too much. We could even specify 1min, it would greatly reduce the dataset to scan.

smarnach: I had to convince Postgres that the datatypes are correct:
```
SELECT age(last_modified, from_epoch((data->'target'->'data'->>'last_modified')::bigint))
FROM objects
WHERE resource_name = 'history'
ORDER BY 1 DESC
LIMIT 100;
```
However, the results are all more than two years, so apparently something isn't quite right. Anyway, this was just meant as a quick verification. I'm also fine with just testing this in production – the worst that can happen is a failing Poucave check, which seems like a risk we can take. :)

leplatrem: > However, the results are all more than two years, so apparently something isn't quite right.

Oh, now I'm curious :) 

> [...]  the worst that can happen is a failing Poucave check, which seems like a risk we can take. :)

No, not even. The only consequence would be that the number of changes for a certain approval would be inaccurate :)

PR #515, opened 2020-07-22
Replace kinto-signer by canonicaljson-rs

leplatrem:
This PR leverages the Canonical JSON serializer in Rust.

Plan is:

* Step 1: replace the serializer of Poucave
* Step 2: if signature validation never fails during a period, it will indirectly make sure that we could replace it in the server
* Step 3: once replaced in the server, we could replace it in the client too :)


mostlygeek (approved):
I like that you remove a layer of abstraction, from `canonical_json(records, timestamp)` to `canonicaljson.dumps(...)`.  

This looks good.  Surprised how small the PR turned out to be :)

PR #566, opened 2020-09-14
Show known bugs in each check

leplatrem:
Screenshot from 2020-09-14 12-42-00


mostlygeek (approved):
I'm approving this even though I'm not sure I entirely understand all of it.  It looks useful and I like that it helps us see more of the picture of a state of a project. 

What use case spurred this?

leplatrem:
> What use case spurred this?

I was working on the blueprint and realized that this part was trivial

PR #569, opened 2020-09-18
Fallback to empty list when bugtracker fetch fails

leplatrem:
See https://github.com/mozilla-services/cloudops-infra/pull/2510


leplatrem:
I'll merge this and roll it out. We can re-iterate next Monday

PR #575, opened 2020-09-25
Send check results to InfluxDB

leplatrem:
The following code works with a local InfluxDB docker. I'm not completely sure about the wording (eg. `org` vs. `db` in InfluxDB world). And whether I used correctly the notions of org/bucket, or whether it's a good idea to default org and bucket from service name.

(*review by commit should be easier*)

```
$ docker run -p 8086:8086 \
      -v $PWD:/var/lib/influxdb \
      influxdb
```

```
make serve -e INFLUXDB_URL=http://localhost:8086 
```

```
curl -G 'http://localhost:8086/query?pretty=true' --data-urlencode "db=poucave" --data-urlencode "q=SELECT \"success\" FROM \"check\"" 
```


leplatrem:
From @mostlygeek: «we loose a lot of information with this approach (eg. *which collection was causing the global error rate to increase?*)»

smarnach:
Oh, I totally missed this review. Looking now.

leplatrem:
Giving up in favor of #603

PR #603, opened 2020-10-08
Output check result data to stdout in Mozlog JSON format

leplatrem:
The idea is that at almost no cost we can send the result data to BigQuery using stdout and the Mozlog JSON format.

I originally thought that only `request.summary` were stored. I didn't want to use the `body` field of the request summary log, because it is **not** representative of the check state change. I think having a dedicated log with just the results when the cache expires is the right way to go, as long as it gets processed and ingested in our BigQuery of course

Based on #601 for the events part


smarnach (commented):
This looks really cool! I haven't reviewed it in detail yet, but I think this is the right approach. We get the a numeric value to plot where it makes sense, we get a Boolean value whether the check succeeded or failed and we get the complete data when we need it, so it's best of all worlds. And it's really easy to get the data into BigQuery this way. As far as I can tell, it's basically better than the other approaches we discussed in every regard – we don't need to run a DB, we can plot the data in Grafana, we persist the full result of every check, and it's eveen simple to do.

smarnach:
I will do a line-by-line review on Monday morning.

poucave/app.py
```diff
@@ -120,22 +128,28 @@ async def run(self, cache=None, force=False) -> Tuple[Any, bool, Any, int]:
...
+                is_check_changed = last_success is not None and last_success != success
+                if is_first_failure or is_check_changed:
+                    events.emit("check:state:changed", payload=payload)
 
         return result
 
+    @property
+    def plot(self):
+        defaul_plot = getattr(self.module, "DEFAULT_PLOT", None)
+        return self._plot or defaul_plot
```
smarnach, line 151: Typo in variable name: `defaul_plot`

poucave/app.py
```diff
@@ -303,18 +329,73 @@ async def _run_checks_parallel(checks, cache, tracker, force=False):
...
+    infos = {
+        "time": utils.utcnow().isoformat(),
+        "project": check.project,
+        "check": check.name,
+        "tags": check.tags,
+        "success": result["success"],
+        # Convert result data to string (for type consistency).
+        "data": json.dumps(result["data"]),
+        # An optional scalar value (see below)
+        "plot": 0.0,
```
smarnach, line 371: Can we use `None` for no data here? It should be translated to `null` in JSON.

smarnach (approved):
This looks good to me!

I'm wondering, where do we configure the logger to log in JSON format? I can't find any code to do so in this PR, so are we already logging in JSON?

poucave/app.py
leplatrem: Since BigQuery may create the column from the data of the first entry I thought that `null` could cause issues. Do we have more info about the code that takes the log entry and process it?

checks/core/cloudfront_age.py
```diff
@@ -8,6 +8,7 @@
 
 
 EXPOSED_PARAMETERS = ["url", "max_age"]
+DEFAULT_PLOT = "."
```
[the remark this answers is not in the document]
leplatrem: You mean like ability to define a function per check module?

leplatrem:
> are we already logging in JSON?

Yes, unless you set `LOG_FORMAT=text` the output is JSON

checks/core/cloudfront_age.py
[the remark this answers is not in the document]
smarnach: Yes, each module could offere a function to extract the data instead of the `DEFAULT_PLOT` string. But we are porbably not going to need it, and if we do, we can still change it. I just mentioned an idea I had when I read the code, no need to change anything.

poucave/app.py
smarnach: This indeed sounds like a valid concern. I don't know what will happen. I think The JSON is simply ingested into StackDriver and Bigtable as it is, and any parsing happens on Google's side. A `null` value may very well result in a table with a string column, but maybe we can simply try it? I don't have a strong opinion either way.

leplatrem: I agree that semantically this `0.0` here is bad.

I'm happy to try, and release a dot version if it fails. But you may have to get your hands dirty too (ie. drop the table) in case it does not get ingested :sweat:

checks/core/cloudfront_age.py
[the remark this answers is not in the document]
leplatrem: Let's keep it this way, it serves as a kind of «interface» between the check and the app :)

poucave/app.py
smarnach: That's fine. Since a new table is created every day, I guess I can also fix it by just waiting for the next day. :)

PR #620, opened 2020-10-23
Return CRLite filter age as floating-point number.

smarnach:
Now that we can plot the filter age, it would be nicer not to round it to full hours.


smarnach:
There's no need to tag a release for this – it's just a cosmetic change.

leplatrem (approved):

PR #659, opened 2020-11-16
Adjust push timestamp comparison to support 5min lag

leplatrem:
Since the push timestamp is now published from lambda, we need to allow
some lag between the source and the destination

### DIFF
--- a/checks/remotesettings/push_timestamp.py
+++ b/checks/remotesettings/push_timestamp.py
@@ -10,7 +10,7 @@ import logging
 import websockets
 
 from poucave.typings import CheckResult
-from poucave.utils import utcfromtimestamp
+from poucave.utils import utcfromtimestamp, utcnow
 
 from .utils import KintoClient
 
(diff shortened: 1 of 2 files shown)


checks/remotesettings/push_timestamp.py
```diff
@@ -48,15 +48,17 @@ async def get_remotesettings_timestamp(uri) -> str:
...
+) -> CheckResult:
     rs_timestamp = await get_remotesettings_timestamp(remotesettings_server)
     push_timestamp = await get_push_timestamp(push_server)
 
     rs_datetime = utcfromtimestamp(rs_timestamp)
     push_datetime = utcfromtimestamp(push_timestamp)
 
     return (
-        push_timestamp == rs_timestamp,
+        0 <= (rs_datetime - push_datetime).seconds < lag_margin,
```
smarnach, line 61: I don't think this is what we want. The difference between these two timestamps can become arbitrarily large – it just can't _remain_ arbitrarily large for longer than `lag_margin`. So what we want is that either `rs_datetime >= datetime.now() - lag_time` or `rs_datetime == push_datetime`.

leplatrem: Oh, of course. It took me some time to understand the difference, but yes of course. Thank you for the feedback

smarnach (approved):
Looks good to me.

There is a theoretical loophole here. If some automated job updates Remote Settings more frequently than every five minutes, the check will never failed, since the timstamp on RS always is recent enough to be within the margin. This seems more like a theoretical risk, though, and I'll get paged when the lambda fails, so this seems acceptable.

PR #664, opened 2020-11-26
Fix #663: Retry fetching of x5u and crash if not signature error

### LINKED ISSUE #663: Retry HTTP requests to make checks more robust
A few days ago, I got paged for a failing Poucave check earlier today. When I looked, everything looked fine, but after some digging in the BigQuery logs I found a failure of the `main-signatures` check with this retur data:

    {"pinning/pins": "ClientOSError(104, 'Connection reset by peer')"}

So apparently a single connection reset was enough to make the check fail, and because of the TTL of two hours, the Pingdom check would fail for the next two hours. We should make that check (and possibly others as well) more robust by adding some retry behaviour when a connection fails. We could also allow setting a different TTL when the check failed, so it can recover more quickly.

leplatrem:
Fix #663

### DIFF
--- a/checks/remotesettings/validate_signatures.py
+++ b/checks/remotesettings/validate_signatures.py
@@ -9,10 +9,16 @@ import time
 from typing import List
 
 import canonicaljson
-from autograph_utils import MemoryCache, SignatureVerifier, decode_mozilla_hash
+from autograph_utils import (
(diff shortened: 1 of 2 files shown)


smarnach (approved):
Thanks for the quick fix!

PR #680, opened 2020-12-08
Add slide out panel with full details of check

rehandalal:
r?


leplatrem:
Oh cool! This looks nice :) 

Some general feedback, before diving in the diff...

* I think we can get rid of the Details + collapse/expand on the card, and move everything except the `Updated X seconds ago` info to the side panel. The `open check API data` link and the refresh buttons could be moved to the panel too (not used that often)
* The panel or its info could be slightly limited in width. Otherwise long descriptions are bit hard to read and info like `Executed in X ms` is too far. But a wide panel helps visualize history :)
* A possible solution would be to show the general info (description, documentation, parameters) in 2 columns. This would also save vertical space to show known issues and troubleshooting link

Screenshot from 2020-12-09 18-04-37

rehandalal:
Sorry about the creep in this PR but I was finding updating components a little tricky with them all in a single file. I decided to take advantage of JS module import/exports and create individual files for each component to make the code a little easier to navigate/maintain. If you hate this I can roll it back to the way it was before without much of a hassle. Also the details panel is now split out into its own component.

poucave/html/js/app/components/CheckDetails.mjs
```diff
@@ -0,0 +1,277 @@
...
+          </div>
+          
+          <hr />
+          
+          <div class="d-flex">
+            <div class="w-50">
+              <h4>Documentation</h4>
+              <${Markdown} md="${data.documentation}" />
+            </div>
+            <div class="pl-2">
```
leplatrem, line 246: Shouldn't it be `w-50` here too?

poucave/html/js/app/components/Check.mjs
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    return html`
+      <div class="card-status card-status-top ${statusClass}"></div>
+      <div class="card-header">
+        <h4 class="card-title check-name">${data.name}</h4>
+        <div class="card-options">
+          <a class="check-url" title="Open check API data" href="${data.url}" target="_blank">
+            <i class="fa fa-sm fa-external-link-alt" />
+          </a>
```
leplatrem, line 74: I believe this can be moved to the details panel too, it's not used that often

poucave/html/js/app/components/Check.mjs
```diff
@@ -0,0 +1,131 @@
...
+          <a class="ml-3" href="#" title="Show more details" onClick=${this.handleToggleDetails}>
+            <i class="fa fa-eye" />
```
leplatrem, line 76: Should we make the whole title (or card?) clickable to open the panel?

leplatrem (approved):
That's cool! Let's hold to our challenge of not having JS build step ;)

r+wc :)

I will have a few things to adjust on the chart (margins etc.), let me know if you prefer to merge this or if I should push to your branch

poucave/html/js/app/components/CheckDetails.mjs
rehandalal: because of flex this will just fill the available space

poucave/html/js/app/components/Check.mjs
rehandalal: 👍🏽

poucave/html/js/app/components/Check.mjs
rehandalal: Yeah if there are no other options I think that makes sense. I'd probably still keep the icon there.

rehandalal:
Lets merge this and then you can tweak

PR #711, opened 2020-12-18
Fix #383: reflect focused in URL and vice-versa 

### LINKED ISSUE #383: Add shareable links for each check
Right now if you click on a failing check on the SVG diagram it jumps you down to the check on that page, but it doesn't add an anchor to the URL so that we can share the link to specific checks.  It would be handy if each check had a shareable link.

leplatrem:
I wanted to do something simple here.

Unfortunately I could not find an elegant way to distinguish animation on context change and details opening/closing without duplicating a lot of code.

By getting rid of the animation, everything remains relatively straightforward. And in terms of UX it makes sense I think, links on the diagram open the details panel without scrolling for example. And a user «browsing» the cards doesn't need scroll. For permalinks, it is more debatable, but not scrolling is not insane.

I was hesitating with using the history state API, but thought that parsing the URL hash was simple enough. Calling  context `setValue()` down the line instead of using simple links in the diagram and checks is debatable. The codebase would be simpler, the only inconvenient is that the URL hash formatting would break DRY, but it's extremely dumb.  What do you think?

### DIFF
--- a/poucave/html/css/main.css
+++ b/poucave/html/css/main.css
@@ -112,22 +112,6 @@ section.project:not(:first-child) {
     max-width: 100%;
 }
 
-@keyframes blinking {
-    0%, 50%, 100% {
-        transform: scale(1, 1);
-        opacity: 1;
-    }
-
-    25%, 75% {
-        transform: scale(0.9, 0.9);
-        opacity: 0.35;
-    }
-}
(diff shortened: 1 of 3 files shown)


rehandalal (approved):
lgtm. r+

PR #768, opened 2021-02-02
Fix #767: check expiration of whole cert chain

### LINKED ISSUE #767: New check for Content Signature certificate chain
Normandy has [a check](https://github.com/mozilla/normandy/blob/dc24516f34d2d9d2920fcb62683b1fe3eca1fda3/normandy/recipes/checks.py#L125-L164) that checks the content signature certificates do not expire soon. 

Duplicating, and expanding this for Poucave would be useful.  It should check Remote Settings content signatures for: 

- signatures are valid 
- certificate chain is valid 
- certificate have at least `x` hours before they expire
- any others?

leplatrem:
Fix #767

### DIFF
--- a/checks/remotesettings/certificates_expiration.py
+++ b/checks/remotesettings/certificates_expiration.py
@@ -23,6 +23,10 @@ logger = logging.getLogger(__name__)
 
 EXPOSED_PARAMETERS = ["server", "min_remaining_days"]
 
+# Bound the alert thresholds.
+LOWER_MIN_REMAINING_DAYS = 7
+UPPER_MIN_REMAINING_DAYS = 60
+
 
 async def fetch_certs(x5u):
     cert_pem = await fetch_text(x5u)
@@ -45,7 +49,11 @@ async def fetch_collection_metadata(server_url, entry):
     return collection["data"]
 
 
-async def run(server: str, min_remaining_days: int) -> CheckResult:
+async def run(
+    server: str,
+    percentage_remaining_validity: int = 10,
+    min_remaining_days: int = LOWER_MIN_REMAINING_DAYS,
+) -> CheckResult:
     client = KintoClient(server_url=server)
     entries = await client.get_monitor_changes()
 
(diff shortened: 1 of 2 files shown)


tests/checks/remotesettings/test_certificates_expiration.py
```diff
@@ -25,6 +26,25 @@
...
+c2lnbmF0dXJlLm1vemlsbGEub3JnMHYwEAYHKoZIzj0CAQYFK4EEACIDYgAE9ft2
+ReAOA01a9iy9RCkXN3v8+6ti5Ba8pjqjq0JYy0D0CgOWFGMWqCCqgPGNC4Ydz1C1
+Ssdh0cjtMHqhGPSpGFttTZqQ+tHa5vfwo7n+KDRqfiY3P4dv/PJitbgzA0hIo4GD
+MIGAMA4GA1UdDwEB/wQEAwIHgDATBgNVHSUEDDAKBggrBgEFBQcDAzAfBgNVHSME
+GDAWgBSgHUoXT4zCKzVF8WPx2nBwp8744TA4BgNVHREEMTAvgi1waW5uaW5nLXBy
+ZWxvYWQuY29udGVudC1zaWduYXR1cmUubW96aWxsYS5vcmcwCgYIKoZIzj0EAwMD
+aQAwZgIxAJFusKXlUSnY8n5ldvqMBUPXZFh0oAitBu1sxpZNGpNuEBSuEWlsAQDE
+gbUgGG/FqgIxANT9km7V/qkriJ99c2MIZsPhKh7suUoPjd7XpbXsRLNle9hxop92
+9eHUYBDGIH8/iA==
+-----END CERTIFICATE-----
```
leplatrem, line 47: This second cert expires after the first one. The datetime shown in the error message in the test below shows that it checks for the one that expires earliest

checks/remotesettings/certificates_expiration.py
```diff
@@ -50,12 +56,15 @@ async def run(server: str, min_remaining_days: int) -> CheckResult:
 
     # Second, deduplicate the list of x5u URLs and fetch them in parallel.
     x5us = list(set(metadata["signature"]["x5u"] for metadata in results))
-    futures = [fetch_cert(x5u) for x5u in x5us]
+    futures = [fetch_certs(x5u) for x5u in x5us]
     results = await run_parallel(*futures)
```
mythmon, line 60: Does this work when `futures` is a list of lists of futures? I believe the type changed from `List<Future<_>>` to `List<List<Future<_>>>`. I'm not very familiar with Python's async tools though, so maybe this is fine?

mythmon (approved):
This is definitely a good change. One thing I have been thinking about this past few days is about the alerting threshold for certs. A certificate that is only valid for 6 weeks (like the signing certs) is fine if it expires in two weeks. A 3 year cert should probably alert when there is a month left though.

Do you think it would make sense to make the alerting be based on a fraction of the total cert life, possibly with some min/max bounds?

checks/remotesettings/certificates_expiration.py
leplatrem: It's a bit simpler than that, `fetch_certs()` used to return `Future<Cert>` and now returns `Future<List<Cert>>`

mythmon: Ahh, right. So the type of `futures` is `List<Future<List<Cert>>>`, so `run_parallel(*futures)` still works fine. Got it.

leplatrem:
> Do you think it would make sense to make the alerting be based on a fraction of the total cert life, possibly with some min/max bounds?

You are totally right, this is a very good idea! 

Currently [we alert if it exprires in 2 weeks](https://github.com/mozilla-services/cloudops-infra/blob/96e1b3554afbf23a6bc7c6fe9a963b43021d7946/projects/poucave/k8s/charts/poucave/configs/poucave/config.toml.tpl#L170).

We could alert if it expires in 10% of the lifetime (6 weeks would be 4 days before, 3 years would be 3 months). Or 5% with a minimum of 7 days, or something like that.

checks/remotesettings/certificates_expiration.py
```diff
@@ -50,23 +65,38 @@ async def run(server: str, min_remaining_days: int) -> CheckResult:
...
-    expirations = {
-        x5u: cert.not_valid_after.replace(tzinfo=datetime.timezone.utc)
-        for x5u, cert in zip(x5us, results)
-    }
+
+    validity: Dict[str, Tuple] = {}
+    for x5u, certs in zip(x5us, results):
+        # For each cert of the chain, keep track of the one that ends the earliest.
```
mythmon, line 73: This is just recording all of the ending times, not finding the one that ends the earliest, right? Is this comment out of date?

mythmon (approved):

checks/remotesettings/certificates_expiration.py
leplatrem: No, we have ` or end < validity[x5u][0]` below

mythmon: Ah, I understand now. :+1:

PR #773, opened 2021-02-19
Remove check about XML timestamp

leplatrem:
For a reason that I don't explain, the XML update timestamp is roughly a
day behind the source collection timestamps.

https://bugzilla.mozilla.org/show_bug.cgi?id=1675807

This XML is used by Firefox < 68, and although it's still live, we don't support it anymore and don't want to spend human resources for this.

Let's remove this failing check that prevents us from concentrating on
real stuff.

### DIFF
--- a/checks/remotesettings/blocked_pages.py
+++ b/checks/remotesettings/blocked_pages.py
@@ -76,8 +76,6 @@ async def run(remotesettings_server: str, blocked_pages: str) -> CheckResult:
         len(missing) == 0
         and len(missing_ids) == 0
         and len(extras_ids) == 0
-        and xml_timestamp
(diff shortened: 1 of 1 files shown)


smarnach (approved):

checks/remotesettings/blocked_pages.py
```diff
@@ -76,8 +76,6 @@ async def run(remotesettings_server: str, blocked_pages: str) -> CheckResult:
         len(missing) == 0
         and len(missing_ids) == 0
         and len(extras_ids) == 0
-        and xml_timestamp
-        in (addons_timestamp, plugins_timestamp, certificates_timestamp)
     )
```
smarnach, line 79: Black thinks the check should be on a single line now.

PR #797, opened 2021-04-15
Fix #795: Add check for state of maintenance

### LINKED ISSUE #795: Add checks about state of maintenance of underlying repositories
Current situation: Remote Settings depends on a certain amount of repositories (kinto, kinto plugins, client sdk and wizard, kinto dist, remote-settings lambdas, ...).  Maintenannce requires time and efforts. Since we rely on Dependabot for all them, it is easy to lose track of ongoing pull-requests.

When everything is fine, all repositories have zero opened pull-requests, tests are green.
When maintainers are under the weather, tests fail, pull-requests stack up, and the overall state of the Remote Settings repositories is preoccupying. However, this is relatively invisible.

In order to force us keep everything clean and green, my idea was to introduce a check that would fail if the number of open pull-requests on the specified repositories is superior to X.

leplatrem:
Fix #795

### DIFF
--- a/checks/core/maintenance.py
+++ b/checks/core/maintenance.py
@@ -25,6 +25,7 @@ logger = logging.getLogger(__name__)
 EXPOSED_PARAMETERS = ["max_days_last_activity", "max_opened_pulls"]
 
 
+@retry_decorator
 async def pulls_info(session, repo):
     token = os.getenv("GITHUB_TOKEN")
     headers = {
@@ -58,7 +59,6 @@ async def pulls_info(session, repo):
     return sorted(pulls)
 
 
-@retry_decorator
 async def run(
     repositories: List[str],
     max_opened_pulls: int = 7,
--- a/checks/core/maintenance.py
+++ b/checks/core/maintenance.py
@@ -1,0 +1,38 @@
+"""
+Inspect a list of Github repositories, the number of opened
+pull requests should be under the specified maximum.
+
+The number of opened pull-requests for each repository is returned.
+"""
+import os
+from typing import List
+
+from poucave.typings import CheckResult
+from poucave.utils import ClientSession, retry_decorator, run_parallel
+
+
+EXPOSED_PARAMETERS = ["max_opened_pulls"]
+
+
+async def count_pulls(session, repo: str) -> int:
+    token = os.getenv("GITHUB_TOKEN")
+    url = f"https://api.github.com/repos/{repo}/pulls"
+    headers = {
+        "Accept": "application/vnd.github.v3+json",
+        "Authorization": f"token {token}" if token else "",
+    }
+    async with session.get(url, headers=headers) as response:
+        data = await response.json()
+    return len(data)
+
+
+@retry_decorator
+async def run(repositories: List[str], max_opened_pulls: int = 10) -> CheckResult:
+    async with ClientSession() as session:
+        futures = [count_pulls(session, repo) for repo in repositories]
+        results = await run_parallel(*futures)
+        zipped = zip(repositories, results)
+        repos_over = {
+            repo: count for (repo, count) in zipped if count > max_opened_pulls
+        }
+        return len(repos_over) == 0, {"pulls": zipped}
--- a/checks/core/maintenance.py
+++ b/checks/core/maintenance.py
@@ -24,7 +24,7 @@ logger = logging.getLogger(__name__)
 
 EXPOSED_PARAMETERS = [
     "max_days_last_activity",
-    "max_days_last_activity",
+    "min_days_last_activity",
     "max_opened_pulls",
 ]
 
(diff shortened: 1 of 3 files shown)


smarnach (commented):
Instead of alerting on the number of pull requests, how about alerting on the age of the oldest open PR instead? We should filter out draft PRs if we decide to go that route.

leplatrem:
> Instead of alerting on the number of pull requests, how about alerting on the age of the oldest open PR instead? 

The number of pull-requests can indeed be annoying when dependabot is scheduled weekly/monthly on a repo, where it's not allowed to auto-merge (eg. kinto-dist).

But I'm not sure the oldest open PR would work. We sometimes have contributors who take their time before taking reviews into account. And it wouldn't catch immediately the dependabot pull-requests that stack up because of failing tests.

We could look into other criterias maybe?
Age of the oldest comment of the pull-request? Age of the oldest unassigned pull-request?
Pull-requests with failing state?

> We should filter out draft PRs if we decide to go that route.

That I will do in any case. Thanks for the suggestion.

smarnach:
I see. My understanding was that you just wanted to avoid that things start rotting, but if you want to get alerted immediately when dependabot PRs pile up, simply using the number seems to be the better measure. Maybe we _also_ want to alert if some PR is inactive for too long, as a separate check? Even if it's a contributor PR, we should take some action on it after a while.

leplatrem:
I adjusted the check to avoid failing too fast: *the number of opened pull requests whose latest activity (comment or update) is older than `ignore_last_days` should be under the specified maximum `max_opened_pulls`*

What do you think?

> Maybe we also want to alert if some PR is inactive for too long, as a separate check? 

I would do everything in the same check. This could be another parameter: `max_days_last_activity = 30` ?

leplatrem:
```toml
[checks.remotesettings.maintenance]
description = ""
module = "checks.core.maintenance"
ttl = 360000
params.repositories = [
    "Kinto/kinto",
    "Kinto/kinto-admin",
    "Kinto/kinto-attachment",
    "Kinto/kinto-changes",
    "Kinto/kinto-signer",
    "Kinto/kinto-emailer",
    "Kinto/kinto-http.js",
    "Kinto/kinto-http.py",
    "Kinto/kinto-wizard",
    "mozilla/remote-settings-devtools",
    "mozilla-services/kinto-amo",
    "mozilla-services/kinto-dist",
    "mozilla-services/poucave",
    "mozilla-services/remote-settings-lambdas",
]
```

```
$ GITHUB_TOKEN=a...bdf LOG_FORMAT=text PYTHONPATH=. poetry run python -m poucave check remotesettings maintenance
{
  "Kinto/kinto-emailer": {
    "pulls": {
      "old": 13,
      "total": 14
    }
  },
  "Kinto/kinto": {
    "pulls": {
      "old": 11,
      "total": 16
    }
  },
  "Kinto/kinto-attachment": {
    "pulls": {
      "old": 5,
      "total": 7
    }
  },
  "Kinto/kinto-http.js": {
    "pulls": {
      "old": 2,
      "total": 3
    }
  },
  "Kinto/kinto-http.py": {
    "pulls": {
      "old": 1,
      "total": 1
    }
  },
  "Kinto/kinto-admin": {
    "pulls": {
      "old": 0,
      "total": 1
    }
  },
  "mozilla-services/poucave": {
    "pulls": {
      "old": 0,
      "total": 1
    }
  },
  "mozilla-services/remote-settings-lambdas": {
    "pulls": {
      "old": 0,
      "total": 1
    }
  }
}
```

smarnach:
Test update time.

checks/core/maintenance.py
```diff
@@ -0,0 +1,95 @@
...
+    async with ClientSession() as session:
+        futures = [pulls_info(session, repo) for repo in repositories]
+        results = await run_parallel(*futures)
+
+        success = True
+        infos = {}
+        for (repo, pulls) in zip(repositories, results):
+            if len(pulls) == 0:
+                continue
+            age_pulls = [(utcnow() - dt).days for dt in pulls]
```
smarnach, line 77: Maybe store `utcnow()` in a variable? Otherwise we are using a slightly different time for each PR, which is kind of weird. I'd even move it out of the outer for loop, so we use the same timestamp for all repos.

checks/core/maintenance.py
```diff
@@ -0,0 +1,95 @@
...
+            old_pulls = [age for age in age_pulls if age > min_days_last_activity]
+            infos[repo] = {
+                "pulls": {
+                    "old": len(old_pulls),
+                    "total": len(pulls),
+                }
+            }
+            if len(old_pulls) > max_opened_pulls:
+                success = False
+            if any([age > max_days_last_activity for age in age_pulls]):
```
smarnach, line 87: Nit: It's sorted, so the oldes one should be the first. You could also omit the square brackets just to make it look cleaner.

smarnach (changes requested):
Looking good, but I still have a few questions. :)

checks/core/maintenance.py
```diff
@@ -0,0 +1,92 @@
...
+    utcfromisoformat,
+    utcnow,
+)
+
+
+logger = logging.getLogger(__name__)
+
+EXPOSED_PARAMETERS = [
+    "max_days_last_activity",
+    "max_days_last_activity",
```
smarnach, line 27: s/max/min/

checks/core/maintenance.py
```diff
@@ -0,0 +1,92 @@
...
+            opened = [
+                utcfromisoformat(p["updated_at"]) for p in pulls if not p["draft"]
+            ]
+            if len(opened) == 0:
+                continue
+            age_pulls = [(now - dt).days for dt in opened]
+            # Fail if too many opened PR.
+            old_pulls = [age for age in age_pulls if age > min_days_last_activity]
+            # Fail if opened PR hasn't received recent activity.
+            if any(age > max_days_last_activity for age in age_pulls):
```
smarnach, line 76: Maybe `if max(age_pulls) > max_days_last_activity:`?

smarnach (approved):
Thanks for the updates!

PR #798, opened 2021-04-15
Replace Redash by BigQuery

leplatrem:

PR #800, opened 2021-04-20
Fix instantiation of BigQuery clients per thread

leplatrem:
Sorry for having merged too quickly :) 

Follow-up of https://github.com/mozilla-services/poucave/pull/798#discussion_r615959719

### DIFF
--- a/poucave/utils.py
+++ b/poucave/utils.py
@@ -18,7 +18,6 @@ from poucave.typings import BugInfo
 
 
 logger = logging.getLogger(__name__)
-threadlocal = threading.local()
 
 
 class Cache:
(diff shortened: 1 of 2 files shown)


poucave/utils.py
```diff
@@ -369,13 +368,14 @@ async def fetch_bigquery(sql):  # pragma: nocover
...
     Execute specified SQL and return rows.
     """
-    bqclient = getattr(threadlocal, "bqclient", None)
-    if bqclient is None:
-        # Reads credentials from env and connects.
-        bqclient = bigquery.Client()
-        setattr(threadlocal, "bqclient", bqclient)
 
     def job():
+        threadlocal = threading.local()
```
smarnach, line 373: This still needs to be a global variable – otherwise you create a new instance of `threading.local` whenever the job is run, so you also create a new BQ client.

leplatrem: Ok, I thought it was the behaviour you intended with the previous comment...

smarnach: Sorry about this! I know this is getting somewhat confusing. The current version doesn't really make sense – we could simply unconditionally create a new BQ client inside `job()` everytime it is run, since this is what's going to happen anyway. Creating a new `thread.local` and then immediately checking whether it has some attribute is just not useful – it will never have that attribute, since it's a new instance.

The behaviour I intended is that whatever thread runs the job uses its own BQ client. There will still be only one BQ client per thread, and they are reused as long as the thread lives. The version currently in master takes the local version of BQ client in the thread that spawns the job, and then sends that BQ client across threads to the thread that runs the job. This can lead to multiple threads using the same BQ client at the same time.

smarnach (approved):
This is what I had in mind. Sorry about the confusing comments!

poucave/utils.py
leplatrem: Now I get it! Thanks!

leplatrem:
>  This is what I had in mind. Sorry about the confusing comments!

Thanks for catching this! Don't be sorry, it's not like things were super clear on my side either ;)

PR #801, opened 2021-04-20
Optimize Uptake queries using check parameters

leplatrem:
With the previous setup, using Redash, the parameterization of cached queries was limited.

Here we can reduce the amount of scanned data using the checks parameters.


leplatrem:
I tested this locally and it seems to work as expected. I'm merging this. Will re-open another follow-up PR to take potential review comments into account.

smarnach:
Looks all good to me, sorry for the delay!

PR #845, opened 2021-08-06
Fix #824: escape now reinitiates the URL bar

### LINKED ISSUE #824: Closing details with Escape doesn't reset URL hash
...and prevents to reopen the same details panel twice

STR

* open check details
* press escape
* details are closed (URL is not reset)
* try to re-open same check details, nothing happens

leplatrem:
See STR in #824


petemoore:
lgtm, thanks!

PR #846, opened 2021-08-06
Fix #796: add background to close button

### LINKED ISSUE #796: Apparence of details close button
Capture d’écran de 2021-04-13 14-40-06

leplatrem:


jwhitlock (approved):
Thanks @leplatrem!

I haven't kept up with CSS (I didn't know ``var(--light)`` was valid), so I looked at how the close button changed.

Before:

poucave_846_before

After:

poucave_846_after

I think this is an improvement, although the white gap below the button bothered me a little. If it bugs you, maybe add a bottom border, like:

```css
border-bottom: 1px solid var(--gray);
```

poucave_846_after_border

Increasing the line height from 50px to 55px also removed the gap, but then the bottom line is really needed or the close button fades into the background:

poucave_846_line_height_55

jwhitlock (approved):
Thanks, 🚢 SHIP IT!

poucave_846_round2

PR #864, opened 2021-08-13
Fix #863: use configured service name in UI

### LINKED ISSUE #863: Pull project name from root URL
Currently, the HTML page title is hardcoded:

https://github.com/mozilla-services/poucave/blob/24fb653571341a8aabe62f9c83716b35dea703b0/poucave/html/index.html#L6

Instead the title should be taken from the root URL endpoint (`/`):

https://github.com/mozilla-services/poucave/blob/24fb653571341a8aabe62f9c83716b35dea703b0/poucave/app.py#L200-L204

Maybe capitalized with dashes removed.

This will be necessary for our FirefoxCI / Taskcluster instance !

leplatrem:
Fix #863

Using `SERVICE_NAME="telescope-firefox-CI"` when running the app, the UI will show `Telescope Firefox CI`.

Note that the service name is also used to pull tickets from Bugzilla whose *whiteboard* field contains the configured string.

### DIFF
--- a/poucave/html/index.html
+++ b/poucave/html/index.html
@@ -3,7 +3,7 @@
 <head>
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
-  <title>Delivery System Status</title>
+  <title></title>
   <link rel="stylesheet" href="css/font-awesome.min.css" />
   <link rel="stylesheet" href="css/tabler.min.css" />
   <link rel="stylesheet" href="css/tabler-dark.min.css" />
--- a/poucave/html/index.html
+++ b/poucave/html/index.html
@@ -3,7 +3,7 @@
 <head>
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
-  <title></title>
+  <title>Loading...</title>
   <link rel="stylesheet" href="css/font-awesome.min.css" />
   <link rel="stylesheet" href="css/tabler.min.css" />
   <link rel="stylesheet" href="css/tabler-dark.min.css" />
--- a/poucave/html/js/app/index.mjs
+++ b/poucave/html/js/app/index.mjs
@@ -1,11 +1,29 @@
 import { html, render } from "../htm_preact.mjs";
 
 import App from "./components/App.mjs";
+import { ROOT_URL } from "./constants.mjs";
 
 // Initialize the app on page load
-window.addEventListener("load", () => {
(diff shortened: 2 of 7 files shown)


poucave/html/index.html
```diff
@@ -3,7 +3,7 @@
 <head>
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
-  <title>Delivery System Status</title>
+  <title></title>
```
jwhitlock, line 6: Does it make sense to have a placeholder, like "Loading configuration..." in case something goes wrong with the configuration request?

poucave/html/js/app/index.mjs
```diff
@@ -1,11 +1,29 @@
...
+window.addEventListener("load", async () => {
   // Clear the application container and render the application
   const appContainer = document.getElementById("app");
   appContainer.innerHTML = "";
-  render(html`<${App} />`, appContainer);
+
+  // Fetch instance metadata.
+  const resp = await fetch(new URL("/", ROOT_URL));
+  const { service } = await resp.json();
+  const title = capitalize(service);
```
jwhitlock, line 15: This worries me. I think it makes sense to use the capitalized service as the title default, but the service is used elsewhere, and it makes sense to me to disconnect the human-visible page title from a service name that may be used in links and APIs.

jwhitlock (approved):
This worked well locally, and I appreciate. It tried a few things to break the title:

```sh
SERVICE_NAME="CTMS"      # Title "CTMS"
SERVICE_NAME="CTMS API"  # Title "CTMS API"
SERVICE_NAME="ctms-api"  # Title "Ctms Api"
```

None of that is too broken, and we're a long way from monitoring CTMS with Poucave.

I'm concerned about manipulating the ``SERVICE_NAME`` for the title. I think the capitalized service name is a decent default, but I can imagine cases where you want them to be different (code names vs product names), or where someone unintentionally breaks code by "fixing" the title.

One solution would be an optional "TITLE_NAME" environment variable, with a default "title" set in the server-side code:

```python
    body = {
        "hello": "poucave",
        "service": config.SERVICE_NAME,
        "environment": config.ENV_NAME,
        "title": config.TITLE_NAME or config.SERVICE_NAME.capitalize()
    }
```

Alternatively, we can address that when we get to a case where the service name and the desired title do not match.

leplatrem:
Thank you @jwhitlock for the super detailed feedback! I made the modifications :)

PR #871, opened 2021-08-26
Fix #851: check latest indexed (part 1)

### LINKED ISSUE #851: [taskcluster] Add check for hooks
If it doesn't exist yet, create a hook that:
- creates a task every X minutes that publishes an artifact, and creates a second task, and makes it dependent on itself, so that 
- creates an other task that depends on the previous task completion or fails

The check will verifies that the second task is always successful.

The first task can only create another task ("decision" tasks) as long as the task has the required scopes

leplatrem:
This runs fine at home, with the config shown in docstring

```
TASKCLUSTER_ROOT_URL=https://community-tc.services.mozilla.com TASKCLUSTER_CLIENT_ID=xxxxxxxxx TASKCLUSTER_ACCESS_TOKEN=xxxxxxxx CONFIG_FILE=taskcluster.toml make check project=queue check=latest-indexed
```

### DIFF
--- a/checks/taskcluster/latest_indexed.py
+++ b/checks/taskcluster/latest_indexed.py
@@ -65,7 +65,7 @@ async def run(
         if getattr(e, "status_code") != 404:
             raise
         # No indexed task found. Failing.
-        return False, f"No task found at {index_path:!r}"
+        return False, f"No task found at {index_path!r}"
 
     # 2. Inspect the task using the queue.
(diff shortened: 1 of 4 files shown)


jwhitlock (approved):
Thanks @leplatrem! The tests are just right - good pytest fixtures, and the mocks are at the appropriate level. I think this part of Taskcluster is stable enough that this is a good testing approach.

PR #872, opened 2021-08-26
 Fix #851: check to create hook (part 2)

### LINKED ISSUE #851: [taskcluster] Add check for hooks

leplatrem:
Based on #871 for utils

### DIFF
--- a/checks/taskcluster/create_hook.py
+++ b/checks/taskcluster/create_hook.py
@@ -1,0 +1,91 @@
+"""
+This check will create a hook with the specified period of execution, if it
+does not exist.
+
+Information about created or existing hook will be returned.
+"""
+import logging
+import textwrap
+
+import taskcluster
+import taskcluster.aio
+import taskcluster.exceptions
+
+from poucave import config
+from poucave.typings import CheckResult
+
+from . import utils as tc_utils
+
+
+logger = logging.getLogger(__name__)
+
+# List which check parameters are visible in the UI.
+EXPOSED_PARAMETERS = ["root_url", "index_path", "max_age"]
+
+TASK_METADATA = {
+    "owner": config.CONTACT_EMAIL,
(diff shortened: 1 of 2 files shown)


petemoore:
The hook definition currently lives [here](https://github.com/mozilla/community-tc-config/blob/a9e6cf97acabeeef55772aa200ee6d43bd3e6479/config/projects/taskcluster.yml#L409-L451), although this is some way from this repo, which I see is not ideal.

At the moment, all entities are managed in that repo, as it is intended to be the source control for all the statically defined entities in the community cluster. If we'd like to manage this hook in the poucave repo instead, we'll need to make sure that tc-admin doesn't delete the hook when it runs, as we'll get race conditions. We should also probably cross-reference the two repos.

I see the benefit of keeping the hook generation close to poucave, but I also see the disadvantage of now the tc-admin repo not having the full (centralised) state for rebuilding the projects. Maybe we should discuss this with the team, to see what we'd like to do. This might be a good idea, but it certainly has caveats, so we should probably document the decision making process for our future selves who forgot the reason we chose one way or the other, and to make sure we capture all the benefits/disadvantages.

leplatrem:
This check ensures that a hook exists.

Whether we want to use it or not can be decided later.
IMO it's fine to have a panel of variety of checks, even if we don't use them all so far. It also helps with decision fatigue.

Beyond that, I get what you say for tc-community, and it's going to be the same for firefox-ci. 
I'll close this PR, and we can revamp it if necessary

PR #875, opened 2021-08-30
Fix #854: check for the Secrets service

### LINKED ISSUE #854: Add check that a secret can be written to taskcluster secrets store and successfully retrieved

leplatrem:
Based on #871

### DIFF
--- a/checks/taskcluster/write_secrets.py
+++ b/checks/taskcluster/write_secrets.py
@@ -47,8 +47,8 @@ async def run(
     try:
         await secrets.get(secret_name)
     except taskcluster.exceptions.TaskclusterRestFailure as e:
-        if getattr(e, "status_code") != 404:
-            raise  # pragma: no-cover
+        if getattr(e, "status_code") != 404:  # pragma: no-cover
+            raise
         return False, f"Secret {secret_name!r} was not stored"
 
     # 2. Remove and check.
@@ -57,7 +57,7 @@ async def run(
         await secrets.get(secret_name)
         return False, f"Secret {secret_name!r} was not removed"
     except taskcluster.exceptions.TaskclusterRestFailure as e:
-        if getattr(e, "status_code") != 404:
-            raise  # pragma: no-cover
+        if getattr(e, "status_code") != 404:  # pragma: no-cover
+            raise
 
     return True, {}
--- a/checks/taskcluster/write_secrets.py
+++ b/checks/taskcluster/write_secrets.py
@@ -4,6 +4,7 @@ A check to verify that the Secrets service is operational.
 Information about the lastest indexed task is returned.
 """
 import logging
+import random
 from datetime import timedelta
 
 import taskcluster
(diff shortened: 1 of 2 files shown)


checks/taskcluster/write_secrets.py
```diff
@@ -0,0 +1,63 @@
...
+from poucave import utils
+from poucave.typings import CheckResult
+
+from . import utils as tc_utils
+
+
+logger = logging.getLogger(__name__)
+
+
+DEFAULT_NAME = "project/taskcluster/secrets-test"
```
petemoore, line 22: This could lead to a conflict if the health check url is called concurrently by e.g. two different users. This could be mitigated by giving the secret a unique name, e.g. if the secret name includes a slugid.

checks/taskcluster/write_secrets.py
```diff
@@ -0,0 +1,63 @@
...
+        "expires": (utils.utcnow() + timedelta(seconds=expires_seconds)).isoformat(),
+        "secret": {"hello": "beautiful world"},
+    }
+    await secrets.set(secret_name, payload)
+    try:
+        await secrets.get(secret_name)
+    except taskcluster.exceptions.TaskclusterRestFailure as e:
+        if getattr(e, "status_code") != 404:  # pragma: no-cover
+            raise
+        return False, f"Secret {secret_name!r} was not stored"
```
petemoore, line 52: What is the reason to handle the two failure cases (404 vs non-404) differently?

checks/taskcluster/write_secrets.py
```diff
@@ -0,0 +1,63 @@
...
+
+    # 2. Remove and check.
+    await secrets.remove(secret_name)
+    try:
+        await secrets.get(secret_name)
+        return False, f"Secret {secret_name!r} was not removed"
+    except taskcluster.exceptions.TaskclusterRestFailure as e:
+        if getattr(e, "status_code") != 404:  # pragma: no-cover
+            raise
```
petemoore, line 61: Note, if the secret name never changes, if one time the check fails to remove the secret, all future checks will fail to create it, so an intermittent/temporary issue would appear as a permanent one. However, this would be solved by using a unique secret name for each test with the above suggestion.

petemoore (changes requested):
Looks great, but I think we'll need unique secret names per test for the reasons described below.

Thanks!

checks/taskcluster/write_secrets.py
leplatrem: The `set()` operation at the beginning of the test does not assert anything whether it exists or not, so I think we're good

checks/taskcluster/write_secrets.py
leplatrem: You're right, not much here

checks/taskcluster/write_secrets.py
petemoore: True, but if two checks run at the same time, and only one can create the secret, unless the requests are identical, there is a good chance one of them won't be able to remove the secret and cause a failure. In the case the request URL and request body (for a secrets.remove there probably is no http body) being identical, normally the idempotent nature of the API design would mean that both requests should succeed, however since they may come from different source IP addresses, I'm not sure whether the idempotency checks in the service will consider the requests to be the same or not, and therefore could still cause a failure.

In general I suspect it will be easier to untangle failures from each other if each secret publish/remove test uses a unique secret name.

PR #877, opened 2021-09-02
Fix wrong default port number in README.md (8080 -> 8000)

petemoore:
I think the default port is [8000](https://github.com/mozilla-services/poucave/blob/f72fef8daea15c5b2b57305a6f8ae058bd829bef/poucave/config.py#L11) rather than 8080.


leplatrem (approved):

leplatrem:
Oops! Thanks!

PR #878, opened 2021-09-02
Avoid having zombie secrets when removing fails

leplatrem:
We introduced a random trailing string in order to avoid conflicts two parallel checks run.

But with this, if the Secrets service fails at removing the secret, we'll have many zombie records! Which is not what we'll want when something is failing...

Plus, from the poucave Web service standpoint, there cannot be two checks running in parallel (we use [a mutex lock](https://github.com/mozilla-services/poucave/blob/da8dbfa3b89b9f6be34e1325ed5b7bb069285ce3/poucave/app.py#L112-L114)). It's probably better to remove this trailing string.

### DIFF
--- a/checks/taskcluster/write_secrets.py
+++ b/checks/taskcluster/write_secrets.py
@@ -4,7 +4,6 @@ A check to verify that the Secrets service is operational.
 Information about the lastest indexed task is returned.
 """
 import logging
-import random
 from datetime import timedelta
 
 import taskcluster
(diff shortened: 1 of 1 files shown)


jwhitlock (approved):
@petemoore caught me up with the conversation outside of this PR. I could see problems either way (unique resources names or continuously creating / deleting the same resource), but at least this avoids a lot of manual cleanup in the failure case.

PR #887, opened 2021-09-13
Add check for certificates expiration

leplatrem:


checks/core/certificate_expiration.py
```diff
@@ -0,0 +1,68 @@
...
+    start = cert.not_valid_before.replace(tzinfo=datetime.timezone.utc)
+    end = cert.not_valid_after.replace(tzinfo=datetime.timezone.utc)
+    lifespan = (end - start).days
+
+    # The minimum remaining days depends on the certificate lifespan.
+    relative_minimum = lifespan * percentage_remaining_validity / 100
+    bounded_minimum = int(
+        min(UPPER_MIN_REMAINING_DAYS, max(min_remaining_days, relative_minimum))
+    )
+    remaining_days = (end - utcnow()).days
```
jwhitlock, line 58: I wanted to make sure I understood this bit.

A Let's Encrypt cert has a 90 day lifespan. That means with the defaults:

* Relative minimum is 9 days
* Bounded minimum is 9 days

We often provision 1-year certs (365 days)

* Relative minimum is 37 days
* Bounded minimum is 37 days

If we pay for a 3-year cert (1095 days)

* Relative minimum is 110 days
* Bounded minimum is 60 days

That seems fine to me, but potentially confusing if the caller configures a smaller window (like 7 days) and the relative minimum is picked instead. But, if they get the alert, they should spend their time replacing the cert, not figuring out the config...

jwhitlock (approved):
This looks good to me, thanks @leplatrem!

checks/core/certificate_expiration.py
leplatrem: Thanks for the prompt review!

> they should spend their time replacing the cert, not figuring out the config...

Hehe true! 

Maybe the default values are not adequate. With `min_remaining_days = 15` and `percentage_remaining_validity = 5`, `UPPER_MIN_REMAINING_DAYS = 30` we have:

* 90 days lifespan: relative is 5 days, bounded is 15 days
* 365 days lifespan: relative is 18 days, bounded is 18 days
* 1095 days lifespan: relative is 55 days, bounded is 30 days

PR #941, opened 2022-01-17
Raise `ValueError` instead of `IndexError` in extract_json()

leplatrem:
We have a situation where we try to extract `.percentiles.75.value` from a dict that is `{"percentiles": "No data"}`

This raises an IndexError just because we try to convert `75` to integer in `extract_json`. For consistency, limit raised exceptions to TypeError and ValueError


matt-boris (approved):
:shipit:

PR #943, opened 2022-01-18
Ignore Telemetry from legacy versions in Remote Settings checks

leplatrem:
The checks now ignore Telemetry from legacy versions. 

In order to include all versions, set the check parameter `include_legacy_versions` to `True`.

```toml
# config-dev.toml
[checks.remotesettings.error-rate]
description = ""
module = "checks.remotesettings.uptake_error_rate"
params.max_error_percentage = 5
params.include_legacy_versions = true
```

```
$ GOOGLE_APPLICATION_CREDENTIALS=key.json LOGGER_DEBUG=debug CONFIG_FILE=config-dev.toml make check project=remotesettings check=error-rate

{
  "sources": {
    "security-state/intermediates": {
      "error_rate": 15.52,
      "statuses": {
        "success": 871,
        "network_error": 49,
        "parse_error": 34,
        "custom_1_error": 31,
        "shutdown_error": 20,
        "offline_error": 19,
        "sync_error": 6,
        "corruption_error": 1
      },
      "ignored": {},
      "min_timestamp": "2022-01-18T09:20:00+00:00",
      "max_timestamp": "2022-01-18T09:30:00+00:00"
    },

...
...
...
`

### DIFF
--- a/checks/remotesettings/uptake_error_rate.py
+++ b/checks/remotesettings/uptake_error_rate.py
@@ -12,6 +12,8 @@ from typing import Dict, List, Optional, Tuple
 from telescope.typings import CheckResult
 from telescope.utils import csv_quoted, fetch_bigquery
 
+from .utils import current_firefox_esr
+
 
 EXPOSED_PARAMETERS = [
     "max_error_percentage",
@@ -45,9 +47,7 @@ WITH uptake_telemetry AS (
       AND event_category = 'uptake.remotecontent.result'
       AND event_object = 'remotesettings'
       AND event_string_value <> 'up_to_date'
-      AND app_version != '69.0'
-      AND app_version != '69.0.1' -- 69.0.2, 69.0.3 seem fine.
-      AND (normalized_channel != 'aurora' OR app_version NOT LIKE '70%')
+      {version_condition}
       {channel_condition}
 )
 SELECT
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by grahamalama — COMMENTED** [review]
I noticed that we added:
```python
include_legacy_versions: bool = False,
```

to many of the function signatures of various `run` functions. We never run `run` where `include_legacy_versions=True` though, except in tests. We include this parameter so that we can do something like:
```python
    if not include_legacy_versions:
        min_version = await current_firefox_esr()
        version_condition = f"AND SAFE_CAST(SPLIT(app_version, '.')[OFFSET(0)] AS INTEGER) >= {min_version[0]}"
```

Since we always seem to want to set `min_version`,  is there harm just assuming the `version_condition` will always be
```python
min_version = await current_firefox_esr()
version_condition = f"AND SAFE_CAST(SPLIT(app_version, '.')[OFFSET(0)] AS INTEGER) >= {min_version[0]}"
```
?

Also, I noticed that we repeat the `version_condition` string three times in this PR. Is it worth extracting to a constant?

leplatrem:
> I noticed that we repeat the version_condition string three times in this PR. Is it worth extracting to a constant?

Good point. I wonder though if it wouldn't add an indirection that would hurt the reading of the SQL query. The query would have a subpart as a constant in utils, which isn't ideal. I'm happy to adjust even if you think it's unacceptable to repeat it. I sense that it could be acceptable in this particular case

grahamalama:
> > I noticed that we repeat the version_condition string three times in this PR. Is it worth extracting to a constant?
> 
> Good point. I wonder though if it wouldn't add an indirection that would hurt the reading of the SQL query. The query would have a subpart as a constant in utils, which isn't ideal. I'm happy to adjust even if you think it's unacceptable to repeat it. I sense that it could be acceptable in this particular case

That's fair reasoning. I was more thinking out loud that suggesting anything prescriptive. LGTM 🚢

PR #961, opened 2022-02-07
 Fix #960: take max_remaining_days into account 

### LINKED ISSUE #960: Parameters to the certifcate_expiration check not working as intended.
The [`certificate_expiration` check](https://github.com/mozilla-services/telescope/blob/335e0df6791c2afeb911c63f7378882a42f002cd/checks/core/certificate_expiration.py) has a few inconsistencies.

* The check takes four parameters, but only two of them are in `EXPOSED_PARAMETERS`. I think the other two should be added as well.
* The `max_remaining_days` doesn't do anything. It's never used in the function body.

leplatrem:


tests/checks/core/test_certificate_expiration.py
```diff
@@ -57,6 +57,20 @@ async def test_negative():
     assert data == {"expires": next_month.isoformat()}
 
 
+async def test_positive_bounded_maximum():
+    url = "https://fake.local"
+
+    last_year = utcnow() - timedelta(days=2365)
```
grahamalama, line 63:
```suggestion
    last_year = utcnow() - timedelta(days=365)
```

Also, I don't know how important it is that this is _exactly_ one year. Like if we need to consider leap years or something. But as long as this just needs to be "about a year in the past", this is probably fine.

tests/checks/core/test_certificate_expiration.py
```diff
@@ -57,6 +57,20 @@ async def test_negative():
...
+    next_month = utcnow() + timedelta(days=15)
```
grahamalama, line 64: Very much a nitpick, but this variable name is only accurate for the 2nd half of any given month. Do we actually want this date to be "a date in the next calendar month", or do we just want the date to be 15 days from now?

grahamalama (approved):

tests/checks/core/test_certificate_expiration.py
leplatrem: Oh weird, your comments showed up when I clicked merge :facepalm: 
Will take them into account in a follow up.

PR #1049, opened 2022-06-27
Leverage kinto-http 10.10

leplatrem:
The code removed here was basically moved to kinto-http 10.10

A BearerToken is instantiated when passed a string that starts with `Bearer`, and same for basic auth when passed a string that contains ":"

Should we allow an empty string to be passed as `auth` and interpret it as None?


grahamalama (approved):
Probably not something to address in this PR, but can we use the `kinto_http.AsyncClient` now that we have that and potentially remove `KintoClient` class from `utils`?

PR #1087, opened 2022-09-21
Ignore pull-requests with blocked label in maintenance check

leplatrem:
The maintenance check reports failure when old pull-requests remain open in a repository.

But sometimes, some dependabot pull-requests are blocked (other dependencies, etc.). 

My proposal here is to ignore the pull-requests that have the `blocked` label.


bsieber-mozilla (approved):
Swapping the check from any open PRs to any non-blocked PRs makes sense to me.

LGTM

PR #1121, opened 2022-11-15
Retry failed requests after 1min

leplatrem:
When the frontend fails to obtain a response from the server (eg. no internet connection, server returns a non-JSON response, eg. 5XX, ...), we don't want to wait for the check TTL, we want to retry the request sooner.

I defined this retry interval arbitrarily to 1min


bsieber-mozilla (approved):
LGTM! 

_(Sorry for the delay!)_

PR #1196, opened 2023-03-29
Make sure mocked endpoint takes at least a few milliseconds to complete

leplatrem:
Sometimes `assert body["duration"] > 0` is false


bsieber-mozilla (approved):
LGTM!

PR #1248, opened 2023-05-31
Mention TROUBLESHOOTING_LINK_TEMPLATE in README

leplatrem:
The default URL used to point to a real wiki page at Mozilla, that is now 404.

We realized that the setting was not documented in the README.

Instead of having such a Mozilla-specific default value, we decided to set its default value to a dummy URL so that its more obvious that it has to be customized


grahamalama (approved):
Can we update the PR description or link to an issue to explain the motivation change?

Also, maybe we should use `example.com` instead of `company.com`, since that's a real site.

PR #1309, opened 2023-09-18
Support commit and abrevation in version numbers

leplatrem:
Before, the deployed version check would compare string strictly.

It now supports versions as output of `git describe --tags --abbrev=4 HEAD~`

### DIFF
--- a/checks/core/deployed_version.py
+++ b/checks/core/deployed_version.py
@@ -20,6 +20,6 @@ async def run(server: str, repo: str) -> CheckResult:
     latest_tag = releases["tag_name"]
 
     return (
-        deployed_version == latest_tag,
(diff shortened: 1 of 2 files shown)


checks/core/deployed_version.py
```diff
@@ -20,6 +20,6 @@ async def run(server: str, repo: str) -> CheckResult:
     latest_tag = releases["tag_name"]
 
     return (
-        deployed_version == latest_tag,
+        deployed_version.startswith(latest_tag),
```
smarnach, line 23: This check passes if version `1.1` is deployed and the latest version is `1.10`. How about
```suggestion
        deployed_version.partition("-")[0] == latest_tag,
```

grahamalama (approved):

PR #1349, opened 2023-11-22
Fix #1320: fix handling of futures in work-in-progress check

### LINKED ISSUE #1320: Signed collection listed in work-in-progress check
Screenshot 2023-10-09 at 12 44 03

leplatrem:
This PR:

* fixes a bug where signed collections would show up in info (because zipping two lists of different sizes)
* does not compare records if status is to-review (no need)
* does not fetch records in parallel anymore (since old wip collections are rare, no need)

### DIFF
--- a/checks/remotesettings/work_in_progress.py
+++ b/checks/remotesettings/work_in_progress.py
@@ -43,45 +43,43 @@ async def run(server: str, auth: str, max_age: int) -> CheckResult:
     ]
     results_metadata = await run_parallel(*futures)
 
-    futures_sources = []
-    futures_destination = []
-    for resource, resp in zip(resources, results_metadata):
-        metadata = resp["data"]
+    too_old = {}
+    for resource, collection_metadata in zip(resources, results_metadata):
+        metadata = collection_metadata["data"]
+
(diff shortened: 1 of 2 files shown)


checks/remotesettings/work_in_progress.py
```diff
@@ -43,45 +43,43 @@ async def run(server: str, auth: str, max_age: int) -> CheckResult:
...
+        metadata = collection_metadata["data"]
+
         # For this check, since we want to detect pending changes,
         # we also consider work-in-progress a pending request review.
         if metadata["status"] not in ("work-in-progress", "to-review"):
             continue
 
-        if last_edit_age(metadata) > max_age:
+        # Ignore collections changed recently.
+        if last_edit_age(metadata) <= max_age:
```
alexcottner, line 56: I like this. Bailout conditions should always come first.

alexcottner:
Longer term - the kinto API should be flexibly enough to return exactly the data that clients need. This app really shouldn't need to have business logic in it. If clients need to do further filtering of data that feels like a miss in our API tier.

alexcottner (approved):

PR #1359, opened 2023-11-29
Follow-up #888: rename Docker image name to mozilla/telescope

leplatrem:
Follow-up #888


grahamalama (approved):
LGTM -- we're just waiting on SRE to create the Dockerhub repo right?

PR #1424, opened 2024-04-24
feat: add HISTORY_PROJECT_ID environment variable

ahoneiser:


telescope/utils.py
```diff
@@ -387,10 +387,17 @@ async def fetch_bigquery(sql):  # pragma: nocover
...
+
         if bqclient is None:
-            # Reads credentials from env and connects.
-            bqclient = bigquery.Client()
+            if config.BIGQUERY_PROJECT_ID:
+                bqclient = bigquery.Client(project=config.BIGQUERY_PROJECT_ID)
+
+            else:
+                # Reads credentials from env and connects.
+                bqclient = bigquery.Client()
```
leplatrem, line 397: I think that `bqclient = bigquery.Client(project=config.BIGQUERY_PROJECT_ID)` would be enough, because when looking at docs https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.client.Client you can see that `None` is the default value

tests/test_utils.py
```diff
@@ -35,6 +36,19 @@ async def test_fetch_bigquery(mock_aioresponses):
...
+async def test_fetch_bigquery_with_specific_project(mock_aioresponses):
+    with mock.patch("telescope.utils.bigquery.Client") as mocked:
+        mocked.return_value.project = "acme-project-id"
+        mocked.return_value.query.return_value.result.return_value = [
+            ("row1"),
+            ("row2"),
+        ]
+        result = await fetch_bigquery("SELECT * FROM {__project__};")
+    mocked.return_value.query.assert_called_with("SELECT * FROM acme-project-id;")
+    assert result == [("row1"), ("row2")]
```
leplatrem, line 49: The mock and this assertion about `result.return_value` is not really necessary for the scope of this test

leplatrem (approved):
Just quick notes about the code that could be simplified

telescope/utils.py
ahoneiser: thanks! simplified this :+1:

tests/test_utils.py
ahoneiser: thanks! I've changed tests in the last commit, this is outdated now..

leplatrem (approved):
Could you. please add a line in https://github.com/mozilla-services/telescope?tab=readme-ov-file#server-configuration ?

ahoneiser:
> Could you. please add a line in https://github.com/mozilla-services/telescope?tab=readme-ov-file#server-configuration ?

oh great hint, done :slightly_smiling_face:

PR #1430, opened 2024-05-13
Move `ruff` and `detect-secrets` to `dev` dependency group

grahamalama:


leplatrem (commented):
May require a INSTALL_DEV target in our makefile for dev operations like lint, test?

grahamalama:
> May require a INSTALL_DEV target in our makefile for dev operations like lint, test?

It seems like doing `poetry install` will install the dev dependencies by default, since that group isn't marked as optional.

So in the future, I would like to make that group optional then modify the Makefile as you suggested

leplatrem (approved):

PR #1441, opened 2024-05-23
Fix #1440: use product details API to filter old clients in spike check

### LINKED ISSUE #1440: Use `current_firefox_esr()` in `uptake_spikes` check instead of relying on configuration value for `min_version`
The `uptake_error_rate` check relies on the product details API to fetch the current ESR number to filter out events from unsupported clients.

https://github.com/mozilla-services/telescope/blob/af3f6de25efa4f1ac2754d262479f2a0ab06eb63/checks/remotesettings/uptake_error_rate.py#L118

We should modify the `uptake_spikes` check to do the same instead of reading the `min_version` from configuration.

Ref https://github.com/mozilla-it/webservices-infra/pull/2058

leplatrem:
Fix #1440

### DIFF
--- a/checks/remotesettings/uptake_spikes.py
+++ b/checks/remotesettings/uptake_spikes.py
@@ -9,7 +9,9 @@ The min/max timestamps give the datetime range of the obtained dataset.
 """
 
 from collections import defaultdict
+from typing import Optional
 
+from checks.remotesettings.utils import current_firefox_esr
 from telescope.typings import CheckResult
 from telescope.utils import fetch_bigquery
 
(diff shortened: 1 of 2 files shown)


alexcottner (approved):
Approving with a small note. I'm unsure if the `current_firefox_esr` method will return exactly what we want. The previous ESR has support until at least version .2 comes out for the current ESR. That's only a few weeks of time every year.

leplatrem:
> The previous ESR has support until at least version .2 comes out for the current ESR. That's only a few weeks of time every year.

Shall we do `-1` and that's it?

alexcottner:
> Shall we do `-1` and that's it?

That would give us the previous release not the previous ESR. I think we'd have to pull something like [this file](https://product-details.mozilla.org/1.0/firefox_history_stability_releases.json), then look for the lowest version that's had a release in the last X months/days.

PR #1458, opened 2024-06-26
Fixes #1457 and #1423 - Resolving bugzilla fetch issues and adding bugzilla ping to heartbeat

### LINKED ISSUE #1423: Add Bugzilla integration check in heartbeat
Like add a `ping()` method to the `BugTracker` class

### LINKED ISSUE #1457: Uncaught exception from telescope - Dictionary KeyError for 'bugs'
An uncaught exception from telescope was recently experienced
```
Uncaught exception:
  File "/app/telescope/middleware.py", line 50, in error_middleware
    response = await handler(request)
               ^^^^^^^^^^^^^^^^^^^^^^
  File "/app/telescope/middleware.py", line 33, in request_summary
    response = await handler(request)
               ^^^^^^^^^^^^^^^^^^^^^^
  File "/app/telescope/utils.py", line 180, in wrapper
    view_result = await func(request)
                  ^^^^^^^^^^^^^^^^^^^
  File "/app/telescope/app.py", line 267, in project_checkpoints
    return await _run_checks_parallel(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/telescope/app.py", line 345, in _run_checks_parallel
    buglist = await tracker.fetch(check.project, check.name)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/telescope/utils.py", line 359, in fetch
    buglist["bugs"],
    ~~~~~~~^^^^^^^^
<class 'KeyError'>
KeyError('bugs')
```

This is likely caused by a specific check not completing successfully.

alexcottner:
- Fixes #1457 - Create empty bugs array with any exception as bugzilla may return a different response than expected
 -  Fixes #1423 - Adding a bugzilla "ping" by querying the `/rest/whoami`

### DIFF
--- a/telescope/utils.py
+++ b/telescope/utils.py
@@ -301,12 +301,14 @@ class BugTracker:
 
     async def ping(self) -> bool:
         """
-        Returns True if we can succesfully hit and parse the /rest/version endpoint.
+        Returns True if we can succesfully hit and parse the /rest/whoami endpoint.
         """
-        url = f"{config.BUGTRACKER_URL}/rest/version"
+        url = f"{config.BUGTRACKER_URL}/rest/whoami"
         try:
-            version = await fetch_json(url)
-            return "version" in version
+            response = await fetch_json(
+                url, headers={"X-BUGZILLA-API-KEY": config.BUGTRACKER_API_KEY}
+            )
+            return "name" in response
         except Exception as e:
             logger.exception(e)
             return False
--- a/telescope/utils.py
+++ b/telescope/utils.py
@@ -334,7 +334,7 @@ class BugTracker:
             if buglist is None:
                 # Fallback to an empty list when fetching fails. Caching this fallback value
                 # will prevent every check to fail because of the bugtracker.
-                default_buglist = {"bugs": []}
+                default_buglist: Dict = {"bugs": []}
                 env_name = config.ENV_NAME or ""
                 url = f"{config.BUGTRACKER_URL}/rest/bug?whiteboard={config.SERVICE_NAME} {env_name}"
                 try:
@@ -345,7 +345,7 @@ class BugTracker:
                 except aiohttp.ClientError as e:
                     logger.exception(e)
                     buglist = default_buglist
-                    
+
                 if self.cache:
                     self.cache.set(cache_key, buglist, ttl=config.BUGTRACKER_TTL)
 
--- a/telescope/utils.py
+++ b/telescope/utils.py
@@ -299,6 +299,19 @@ class BugTracker:
     def __init__(self, cache=None):
         self.cache = cache
 
+    async def ping(self) -> bool:
+        """
+        Returns True if we can succesfully hit and parse the /rest/version endpoint.
+        """
+        url = f"{config.BUGTRACKER_URL}/rest/version"
+        try:
+            version = await fetch_json(url)
+            return "version" in version
+        except Exception as e:
+            print(e)
+            logger.exception(e)
+            return False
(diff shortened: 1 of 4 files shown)


telescope/utils.py
```diff
@@ -299,6 +299,19 @@ class BugTracker:
...
+    async def ping(self) -> bool:
+        """
+        Returns True if we can succesfully hit and parse the /rest/version endpoint.
+        """
+        url = f"{config.BUGTRACKER_URL}/rest/version"
+        try:
+            version = await fetch_json(url)
+            return "version" in version
+        except Exception as e:
+            print(e)
```
leplatrem, line 311: leftover?

telescope/utils.py
```diff
@@ -299,6 +299,19 @@ class BugTracker:
     def __init__(self, cache=None):
         self.cache = cache
 
+    async def ping(self) -> bool:
+        """
+        Returns True if we can succesfully hit and parse the /rest/version endpoint.
+        """
+        url = f"{config.BUGTRACKER_URL}/rest/version"
```
alexcottner, line 306: This seemed like the best endpoint for a "ping".

leplatrem, line 306: I would suggest to use a endpoint that verifies the provided API key `config.BUGTRACKER_API_KEY`. Like this one? https://github.com/mozilla/jira-bugzilla-integration/blob/3cd50288b468cc76e0a1c42884368d6ef4e6d577/jbi/bugzilla/client.py#L65-L72

telescope/utils.py
```diff
@@ -324,7 +337,7 @@ async def fetch(self, project: str, name: str) -> List[BugInfo]:
                     buglist = await fetch_json(
                         url, headers={"X-BUGZILLA-API-KEY": config.BUGTRACKER_API_KEY}
                     )
```
leplatrem, line 339: You may want to add an assertion here like `assert "bugs" in buglist` ?

telescope/utils.py
alexcottner: Ah that's a good call. Adjusting

telescope/utils.py
alexcottner: ```Use of assert detected. The enclosed code will be removed when compiling to optimised byte code```
That seems to not be useful or am I misunderstanding the lint error?

leplatrem: Yeah, when your run the app with `python -O`, assertions are not executed, which can be misleading

leplatrem: We can ignore the rule I think since we don't run with `-O` or `PYTHONOPTIMIZE=1`

telescope/utils.py
```diff
@@ -324,7 +338,7 @@ async def fetch(self, project: str, name: str) -> List[BugInfo]:
...
-                except aiohttp.ClientError as e:
+                except Exception as e:
```
leplatrem, line 341: I don't think fetch can raise anything else than `ClientError`, so if we don't raise anything when `"bugs" is not in buglist`, then #1457 is not fixed by this PR

```py
try:
     buglist = ...
     if "bugs" not in buglist:
         raise ValueError(...)
except (aiohttp.ClientError, ValueError) as e:
     ...
```

telescope/utils.py
alexcottner: Ended up adjusting to handle it another way. What do you think?

leplatrem (approved):

PR #1488, opened 2024-09-10
Drop taskcluster checks

leplatrem:
The original idea was to deploy a Telescope instance that would monitor high level stuff on a taskcluster service. Was never deployed. Dropping it.


alexcottner (approved):
:drooling_face:

PR #1497, opened 2024-10-16
Modify `build-publish` workflow to push to GAR

grahamalama:
Maintains current behavior of pushing to Dockerhub, but additionally pushes to GAR


alexcottner (approved):
Looks correct

PR #1516, opened 2024-12-03
Fix #1422: add BigQuery check to heartbeat

### LINKED ISSUE #1422: Add BigQuery integration check in heartbeat
For example call `fetch_bigquery()` to check that credentials configuration is correct

leplatrem:
Fix #1422

### DIFF
--- a/telescope/app.py
+++ b/telescope/app.py
@@ -216,6 +216,7 @@ async def lbheartbeat(request):
 @routes.get("/__heartbeat__")
 async def heartbeat(request):
     checks = {}
+
     # Check that `curl` has HTTP2 and HTTP3 for `checks.core.http_versions`
     curl_cmd = subprocess.run(
         [config.CURL_BINARY_PATH, "--version"],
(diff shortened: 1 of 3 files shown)


telescope/app.py
```diff
@@ -228,8 +229,17 @@ async def heartbeat(request):
...
         else f"missing features {', '.join(missing_features)}"
     )
+
+    # Bugzilla
     bz_ping = await request.app["telescope.tracker"].ping()
     checks["bugzilla"] = "ok" if bz_ping else "Bugzilla ping failed"
+
+    # Big Query
+    try:
+        checks["bigquery"] = (await utils.fetch_bigquery("SELECT 'ok';"))[0]
```
alexcottner, line 239: If bigquery is having an outage, this will also fail for us. Do we want this in our heartbeat because of that? Is there a way for us to check the exception to see if we received an auth error or a different error?

leplatrem:
Superseded by https://github.com/mozilla-services/telescope/pull/1719

PR #1523, opened 2024-12-19
Fixing telescope push timestamp check

alexcottner:
We updated the kinto client library and started using the changeset endpoint.
However that endpoint returns records sorted by bucket and collection name, causing this check to be a false positive.
Sorting the records by `last_modified` first makes this check work and prevents us from hitting this banana peel in the future.

### DIFF
--- a/checks/remotesettings/push_timestamp.py
+++ b/checks/remotesettings/push_timestamp.py
@@ -10,10 +10,8 @@ import logging
 
 import websockets
 
-from operator import attrgetter
 from telescope.typings import CheckResult
 from telescope.utils import utcfromtimestamp, utcnow
-from datetime import datetime
 
 from .utils import KintoClient
 
--- a/checks/remotesettings/push_timestamp.py
+++ b/checks/remotesettings/push_timestamp.py
@@ -10,8 +10,10 @@ import logging
 
 import websockets
 
+from operator import attrgetter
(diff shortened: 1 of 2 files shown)


leplatrem (approved):
Good catch. This regression was introduced in #1515 

The fix would work, but I'd like to contrast it with other solutions:

* Switch back to `get_records()` in the check code since it's what's used in the lambda https://github.com/mozilla-services/remote-settings-lambdas/blob/9e32cf6d9a4d6f9dec5d3d583c3f973c3184a51a/commands/sync_megaphone.py#L53-L62

* Use the changeset endpoint in the sync megaphone job too https://github.com/mozilla-services/remote-settings-lambdas/pull/1506 Since it's the endpoint that is used by clients to poll for changes.

* Sort entries explicitly by `-last_modified` in the changeset endpoint, since there's no reason the list not to be sorted. And align both records and changeset https://github.com/mozilla/remote-settings/pull/725

I approve the PR anyway in case we want the check to be green before our holidays :)

alexcottner:
I vote we make this green before the holidays and then follow up. 

If we modify how `changeset` sorts things, there might be a few other impacts to think about. Like the remote-settings-devtools. But that would be a clean solution if we make everything use changeset and it sorts things in an expected way.

PR #1545, opened 2025-03-10
Fix #1543: add check to verify age of resource

### LINKED ISSUE #1543: Add a resource max age check
The check would take a URL and verify that the response `Last-Modified` response header is less than X hours.

Context: we have a cronjob that is supposed to publish an archive every day. We want to be alerted when the file is too old.

See also https://github.com/mozilla-services/telescope/blob/main/checks/core/cloudfront_age.py

leplatrem:
Fix #1543

See how it will be used in https://github.com/mozilla-it/webservices-infra/pull/4207

### DIFF
--- a/checks/core/resource_age.py
+++ b/checks/core/resource_age.py
@@ -15,7 +15,10 @@ DEFAULT_PLOT = ".age_hours"
 async def run(url: str, max_age_hours: int) -> CheckResult:
     _, headers = await fetch_head(url)
 
-    last_modified = headers.get("Last-Modified", "Mon, 01 Jan 1970 00:00:00 GMT")
+    try:
+        last_modified = headers["Last-Modified"]
+    except KeyError:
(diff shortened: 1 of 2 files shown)


alexcottner (approved):
Seems simple enough

checks/core/resource_age.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+EXPOSED_PARAMETERS = ["url", "max_age_hours"]
+DEFAULT_PLOT = ".age_hours"
+
+
+async def run(url: str, max_age_hours: int) -> CheckResult:
+    _, headers = await fetch_head(url)
+
+    last_modified = headers.get("Last-Modified", "Mon, 01 Jan 1970 00:00:00 GMT")
```
say-yawn, line 18: Can the `last_modified` date header actually be empty or non-existent? If it should never be empty maybe it's worth throwing an error?

PR #1581, opened 2025-08-04
Fix mozilla/remote-settings#951: Check that all source collections are in monitor/changes

leplatrem:
https://github.com/mozilla/remote-settings/issues/951

### DIFF
--- a/checks/remotesettings/collections_consistency.py
+++ b/checks/remotesettings/collections_consistency.py
@@ -104,7 +104,11 @@ async def has_inconsistencies(server_url, auth, resource):
 
 
 async def run(server: str, auth: str) -> CheckResult:
-    resources = await fetch_signed_resources(server, auth)
+    try:
+        resources = await fetch_signed_resources(server, auth)
+    except ValueError as e:
+        msg, affected_collections = e.args
+        return False, {f"{bid}/{cid}": msg for bid, cid in affected_collections}
 
     futures = [has_inconsistencies(server, auth, resource) for resource in resources]
     results = await run_parallel(*futures)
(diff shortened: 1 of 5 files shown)


checks/remotesettings/collections_consistency.py
```diff
@@ -104,7 +104,11 @@ async def has_inconsistencies(server_url, auth, resource):
 
 
 async def run(server: str, auth: str) -> CheckResult:
-    resources = await fetch_signed_resources(server, auth)
+    try:
+        resources = await fetch_signed_resources(server, auth)
+    except ValueError as e:
```
smarnach, line 109: Can we maybe make this a more specific exception? `fetch_signed_resources()` already explicitly throws two different `ValueError`s before this PR, and the PR adds a third one. In addition, virtually everything built into Python might implicitly throw a `ValueError`, so if we catch `ValueError`, we don't really know what we will get.

tests/checks/remotesettings/test_collections_consistency.py
```diff
@@ -4,7 +4,9 @@
 
 
 FAKE_AUTH = "Bearer abc"
-COLLECTION_URL = "/buckets/{}/collections/{}"
+COLLECTIONS_URL = "/buckets/{}/collections"
+COLLECTION_URL = COLLECTIONS_URL + "/{}"
+CHANGESET_URL = "/buckets/{}/collections/{}/changeset"
```
smarnach, line 9: Maybe
```suggestion
CHANGESET_URL = COLLECTION_URL + "/changeset"
```
to be consistent with the previous and following line?

PR #1597, opened 2025-10-03
Speed up tests (from 20sec to 3sec)

leplatrem:


alexcottner (approved):
Seems to work reliably for me. Nice work.

Issue #1088, opened 2022-09-22
Event loop got stuck when requests timed out


leplatrem:
From [Slack](https://mozilla.slack.com/archives/CL7SGADH9/p1663812263304629) 

> We apparently had an issue with the name service in the Kubernetes cluster for Telescope. The name service returned stale IP addresses for [settings.prod.mozaws.net](http://settings.prod.mozaws.net/) inside the Telescope pod, which somehow caused the whole service to stall and not respond to requests anymore. We could fix it after a lot of debugging by getting the name service fixed, but Telescope shouldn't lock up just because it times out on a URL. It kind of looks like it was doing blocking calls to establish connections, eventually stalling the event loop because all threads were blocking.
> 
> Here is one of the error we saw in Sentry: https://sentry.prod.mozaws.net/operations/poucave-prod/issues/18284097/ (edited) 
> 
> The traceback doesn't really look like the code should be blocking, but Telescope was completely unreachable, so something must have been blocking.

leplatrem:
I tried to simulate a slow server with:
```
docker run --publish 0.0.0.0:8888:8888 alpine nc -lk -p 8888 -e sleep 60s
```
([source](https://carltonhenderson.github.io/simulating-slow-http-servers/))

Using:

```toml
[checks.test.hb]
description = ""
module = "checks.core.heartbeat"
params.url = "http://localhost:8888"
```

I obtain consistent timeout exceptions, which are retried via backoff.

Next step would consist in identifying whether the event loop is actually blocked during the connection or not. Or whether the issue comes from the only thread executor being overloaded when many calls are timeout'ing

PR #1724, opened 2026-05-28
Ref #1088: Use aiodns so that DNS lookup don't block threads

leplatrem:
Ref #1088


telescope/utils.py
```diff
@@ -333,9 +333,23 @@ def client_session_start(
...
+    )
+    # Use aiodns-backed AsyncResolver so DNS does not consume the default
+    # ThreadPoolExecutor (shared with BigQuery and other run_in_executor work),
+    # and cache results so repeat lookups don't re-hit a slow nameserver.
+    resolver_loop = loop or asyncio.get_event_loop()
+    connector = aiohttp.TCPConnector(
+        resolver=aiohttp.AsyncResolver(loop=resolver_loop),
+        use_dns_cache=True,
+        ttl_dns_cache=config.REQUESTS_DNS_CACHE_TTL_SECONDS,
+    )
```
leplatrem, line 348: https://stackoverflow.com/a/45169094

uv.lock
```diff
@@ -6,6 +6,18 @@ resolution-markers = [
     "python_full_version < '3.13'",
 ]
 
+[[package]]
+name = "aiodns"
+version = "3.6.1"
```
alexcottner: Version 4.0.4 is the current one. Why start with 3?

alexcottner (approved):
Just have a question about using v3 vs v4.

uv.lock
leplatrem: Oh good catch, I had started this branch a while ago when we were investigating timeouts

